PR #144, opened 2026-02-06
feat: improve key deletion

walterchris:
Properly pass over key ownership and only delete keys if its owned by us.


mhatrevi (approved):
:shipit:

PR #186, opened 2026-02-18
New test cases for AES CBC and fixes for uncovered issues

kenneth-tucker:


api/native/src/algo/aes/cbc.rs
```diff
@@ -54,20 +47,32 @@ impl AesCbcEncryptContext {
         Self {
             context: ctx,
             params: params as *mut AzihsmAlgoAesCbcParams,
+            is_finished: false,
```
mhatrevi: Let's remove this from native layer. This is something to be handled at the RUst API layer; we can do that as a separate task.

api/tests/cpp/utils/known_answer_tests.cpp
```diff
@@ -0,0 +1,269 @@
+// Copyright (c) Microsoft Corporation.
```
mhatrevi: We should not have tests in the utils folder. These tests belong in their individual folders. Add the cbc KATs to cbc.rs, GCM KATs to gcm.rs ...

mhatrevi (changes requested):
Please see the feedback.

api/tests/cpp/utils/known_answer_tests.cpp
kenneth-tucker: I moved the CBC specific content to the cbc test file, but I kept the common utility here. I did rename the file to 'known_answer' to make it clear this is just to support known answer testing, without containing the actual known answer test cases.

api/native/src/utils.rs
```diff
@@ -66,6 +85,13 @@ pub(crate) fn validate_output_buffer(
     output_buf: &mut crate::AzihsmBuffer,
     required_len: usize,
 ) -> Result<&mut [u8], AzihsmStatus> {
+    if output_buf.ptr.is_null() {
+        // Only allow null buffer if length is 0 (for size-query case)
+        if output_buf.len != 0 {
+            Err(AzihsmStatus::InvalidArgument)?;
+        }
+    }
```
mhatrevi: Combine this in a single check. if output_buf.ptr.is_null() &&  output_buf.len != 0

mhatrevi (approved):

PR #231, opened 2026-03-03
[test] expand xts test cases  add more edge and negative tests.

meilu-git:


api/tests/src/algo/aes/xts_tests.rs
```diff
@@ -421,3 +421,324 @@ fn aes_xts_tweak_overflow_rejected(session: HsmSession) {
...
+    let tweak = [0u8; AES_XTS_TEST_TWEAK_SIZE];
+    let dul = 128;
+
+    let algo = HsmAesXtsAlgo::new(&tweak, dul).unwrap();
+    let mut ctx = algo.decrypt_init(key).unwrap();
+
+    let empty: [u8; 0] = [];
+    let out_len = ctx.update(&empty, None).unwrap();
+    assert_eq!(out_len, 0);
+}
```
rajesh-gali: can we add a streaming test to validate 1 dul per update? looks like all of the streaming tests are using multiple dul per update

PR #232, opened 2026-03-03
API integration test

jenstopp:
Add a provider-tests crate that exercises the azihsm provider through
the OpenSSL EVP API (not the CLI tool), enabling testing of session-based
ephemeral keys that cannot be tested via the command line.

Tests added:
- SmokeTest: provider loads and reports correct name
- EC session key sign/verify round trip (P-256, P-384, P-521)
- RSA session key sign/verify (PKCS#1 v1.5 and PSS padding)
- RSA session key encrypt/decrypt (OAEP)
- ECDH session key derivation (to buffer and to file)
- Negative tests: cross-key failures, tampered data, mismatched curves

Additional coverage added in follow-up commits:

- Digest streaming tests (SHA-256, SHA-384, SHA-512)
- HMAC tests (SHA-256, SHA-384, SHA-512) using ECDH-derived masked keys
- HKDF key derivation tests (with/without salt and info parameters)
- ECDH → HKDF → HMAC end-to-end round trip

### DIFF
--- a/plugins/ossl_prov/integration-tests/openssl-capi/build.rs
+++ b/plugins/ossl_prov/integration-tests/openssl-capi/build.rs
@@ -3,33 +3,26 @@
 
 //! Build script for the OpenSSL C API provider integration tests.
 //!
-//! Compiles the C++ GoogleTest test binary via CMake. Requires `OPENSSL_DIR`
-//! to be set, pointing to an OpenSSL 3.0.3 installation prefix (e.g.
-//! `/opt/openssl-3.0.3`). The system OpenSSL is never used.
+//! Compiles the C++ GoogleTest test binary via CMake only when the
+//! `integration` feature is active. This avoids downloading/building
+//! googletest during normal `cargo build` or `cargo clippy` workspace runs.
 //!
-//! When `OPENSSL_DIR` is not set the cmake build is skipped so that
-//! `cargo clippy --all-targets` can check the Rust source without requiring
-//! an OpenSSL installation.
+//! When `integration` is enabled, `OPENSSL_DIR` must point to an OpenSSL 3.x
+//! installation prefix (e.g. `/opt/openssl-3.0.3`).
 
 fn main() {
     println!("cargo::rerun-if-env-changed=OPENSSL_DIR");
 
-    let openssl_dir = match std::env::var("OPENSSL_DIR") {
-        Ok(dir) => dir,
-        Err(_) => {
-            if std::env::var("CARGO_FEATURE_INTEGRATION").is_ok() {
-                panic!(
-                    "OPENSSL_DIR must be set when building with the `integration` feature. \
-                     Point it to your OpenSSL 3.x install prefix."
-                );
-            }
-            println!(
-                "cargo:warning=OPENSSL_DIR is not set — skipping C++ test build. \
-                 Set OPENSSL_DIR to an OpenSSL 3.x installation prefix to build the gtest binary."
-            );
-            return;
-        }
-    };
+    if std::env::var("CARGO_FEATURE_INTEGRATION").is_err() {
+        return;
+    }
+
+    let openssl_dir = std::env::var("OPENSSL_DIR").unwrap_or_else(|_| {
+        panic!(
+            "OPENSSL_DIR must be set when building with the `integration` feature. \
+             Point it to your OpenSSL 3.x install prefix."
+        );
+    });
 
     cmake::Config::new("cpp")
         .define("OPENSSL_ROOT_DIR", &openssl_dir)


xtask/src/coverage.rs
```diff
@@ -30,7 +30,7 @@ impl Xtask for Coverage {
         log::info!("Building all tests and running them with coverage");
         cmd!(
             sh,
-            "cargo llvm-cov nextest --no-report --no-fail-fast --features mock --profile ci-mock"
+            "cargo llvm-cov nextest --no-report --no-fail-fast --features mock --profile ci-mock --workspace --exclude integration-tests"
         )
```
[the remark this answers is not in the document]
jenstopp: fixed

provider-integration-tests/openssl-capi/build.rs
rajesh-gali: I think we should move the tests to plugins/ossl_prov , as integration tests are specific to openssl provider

.github/workflows/rust.yml
```diff
@@ -149,13 +149,19 @@ jobs:
...
         OPENSSL_BIN: /opt/openssl-3.0.3/bin/openssl
         OPENSSL_LIB: ""
         PROVIDER_PATH: /usr/lib/ossl-modules
         PROPQUERY: "?provider=azihsm"
-      run: cargo xtask precheck --nextest --package integration-tests --features integration --profile ci-provider-integration
+      run: cargo xtask precheck --nextest --package provider-integration-tests-cli --features integration --profile ci-provider-integration
+
+    - name: Run C API integration tests
+      env:
+        OPENSSL_DIR: /opt/openssl-3.0.3
```
rajesh-gali: Openssl installation instructions in readme is different than ci. 
Readme : OPENSSL_VERSION=3.0.16
Installation path is : OPENSSL_DIR=/opt/openssl-static

plugins/ossl_prov/integration-tests/openssl-capi/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[dev-dependencies]
+libtest-mimic.workspace = true
+
+[lints]
+workspace = true
+
+[[test]]
+harness = false
+name = "capi_gtest_runner"
```
rajesh-gali: can we rename the binary to something "openssl_capi_integration_tests" to align with other executables.

jenstopp:
@rajesh-gali here I tried to write up the problem I was facing during the development of the test suite (CLI and CAPI) with usage of the config file

# Provider double-init / deadlock: the core problem and how the test suites avoid it

## What `activate = 1` does in `openssl.cnf`

The config has `activate = 1` in the `[azihsm_sect]` section. This tells libcrypto to **load and initialize the provider immediately during config processing** — the application never calls `OSSL_PROVIDER_load()` itself. The internal call chain inside libcrypto:

```
OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CONFIG, ...)
  → ossl_init_config_settings()                        // RUN_ONCE macro
    → ossl_config_int()
      → CONF_modules_load_file()                       // reads OPENSSL_CONF env var
        → provider_conf_load()                         // checks activate=1
          → ossl_provider_activate()
            → provider_init()
              → DSO_load() + DSO_bind_func("OSSL_provider_init")
                → provider's OSSL_provider_init()      // entry point in azihsm_provider.so
```

Config loading is triggered either by an explicit `OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CONFIG, ...)` call, or implicitly when provider operations (e.g. `ossl_provider_find()`) trigger config loading internally. The `openssl` CLI binary calls it explicitly during startup.

Without `activate = 1`, the section just *declares* the provider — it's only loaded when explicitly requested. With it, `provider_conf_load()` checks the `activate` key and calls `ossl_provider_activate()` to load the provider immediately. This is the standard way to make a provider available to CLI tools like `openssl genpkey`, which don't have their own code to call `OSSL_PROVIDER_load()`.

## The core problem

When OpenSSL's default library context auto-loads this config (triggered by the `OPENSSL_CONF` env var), it loads and initializes the azihsm provider into the default context.

The deadlock occurs when the provider's code uses OpenSSL API functions internally (`EVP_MD_fetch`, crypto primitives, etc.). If the provider embeds its own copy of libcrypto (static linking), **that second libcrypto instance auto-initializes on first use** — sees `OPENSSL_CONF` in the environment, reads the config, encounters `activate = 1`, and attempts to load the provider again. Recursive init, deadlock.

The provider doesn't explicitly call OpenSSL initialization — it's the **auto-init mechanism** in the statically-linked libcrypto copy that triggers it. Any OpenSSL API call from the provider is enough to start the chain.

## How each suite avoids it

### CLI tests — shared OpenSSL (single libcrypto)

With shared linking, the `openssl` binary and `azihsm_provider.so` share the **same** `libcrypto.so.3` via the dynamic linker. The dynamic linker loads it once. When the binary's libcrypto reads `openssl.cnf` and loads the provider, the provider's calls into libcrypto go to the already-initialized single instance — no second auto-init, no recursion.

This is why the CI deadlocked before: OpenSSL was built with `no-shared -fvisibility=hidden`, creating **two separate static copies** of libcrypto. The fix was building OpenSSL with shared libraries (`.github/workflows/rust.yml`).

### CAPI tests — `OPENSSL_INIT_NO_LOAD_CONFIG` + explicit context

The CAPI tests use a two-layer defense:

1. **`OPENSSL_init_crypto(OPENSSL_INIT_NO_LOAD_CONFIG, NULL)`** (`plugins/ossl_prov/integration-tests/openssl-capi/cpp/main.cpp:13`) — suppresses automatic config loading on the default context. This skips the config loading step entirely (runs `ossl_init_no_config()` instead, which marks config as "done" without loading any file), so even though `OPENSSL_CONF` is set, the default context never reads it and the provider is never loaded into the default context.

2. **`OSSL_LIB_CTX_new()` + `OSSL_LIB_CTX_load_config()`** (`plugins/ossl_prov/integration-tests/openssl-capi/cpp/utils/provider_ctx.hpp:25-44`) — each test creates a **dedicated, non-default** library context and explicitly loads the config into it. The provider loads into this isolated context only. Since the default context never loaded the provider, there's no double-init.

## Summary

| Suite | Strategy | Why no double-init |
|-------|----------|-------------------|
| CLI | Shared libcrypto | One `libcrypto.so.3` instance; provider calls re-enter the same already-initialized library — no second auto-init possible |
| CAPI | `NO_LOAD_CONFIG` + per-test `OSSL_LIB_CTX` | Default context never loads provider; config loaded explicitly into isolated contexts — safe regardless of static/shared linking |

The CAPI approach is inherently safe regardless of static/shared linking because it never lets the default context auto-load the provider. The CLI approach depends on shared linking to avoid the recursive init.

---

## Annotated call stacks

### CLI test suite

```
1. Rust test runner (nextest) invokes bash script
   e.g. plugins/ossl_prov/integration-tests/openssl-cli/testfiles/ec/round_trip/round_trip.sh

2. Script sources env.sh
   plugins/ossl_prov/integration-tests/openssl-cli/testfiles/env.sh
   ├── Generates key material in $REPO_ROOT/target/test-keymat/cli/     (lines 59-87)
   ├── Generates openssl.cnf with activate=1 for azihsm provider        (lines 95-119)
   │   └── module = <absolute path to azihsm_provider.so>
   │   └── activate = 1
   └── export OPENSSL_CONF="$AZIHSM_KEY_DIR/openssl.cnf"                (line 121)

3. Script calls $OPENSSL_BIN (e.g. genpkey, dgst)
   └── openssl binary starts
       └── libcrypto.so.3 config loading (OPENSSL_INIT_LOAD_CONFIG)
           └── reads OPENSSL_CONF
               └── activate=1 → OSSL_PROVIDER_load("azihsm")
                   └── dlopen(azihsm_provider.so)
                       └── OSSL_provider_init()
                           └── provider uses EVP_* calls → same libcrypto.so.3
                               (shared linking: single instance, already initialized,
                                no recursion)

4. openssl CLI executes the cryptographic operation (genpkey/dgst/req)
   using the now-loaded azihsm provider via -propquery "?provider=azihsm"
```

**Key requirement**: OpenSSL must be built with shared libraries so the binary and provider share one `libcrypto.so.3`. CI builds OpenSSL with `./Configure --prefix=/opt/openssl-3.0.3 --libdir=lib -fPIC` (`.github/workflows/rust.yml:122-123`).

### CAPI test suite

```
1. Rust test runner discovers gtest binary and enumerates tests
   plugins/ossl_prov/integration-tests/openssl-capi/cpp/openssl_capi_integration_tests.rs
   ├── Generates key material in $REPO_ROOT/target/test-keymat/capi/     (generate_dev_key_material())
   ├── Generates openssl.cnf with activate=1                             (generate_openssl_conf())
   └── Calls gtest binary with --gtest_list_tests, parses output         (parse_gtest_list())
       └── Passes OPENSSL_CONF to each subprocess via .env()             (lines 336, 408)

2. For each test, Rust runner invokes the gtest binary with --gtest_filter=Suite.Test

3. gtest binary main() runs BEFORE any test
   plugins/ossl_prov/integration-tests/openssl-capi/cpp/main.cpp:13
   └── OPENSSL_init_crypto(OPENSSL_INIT_NO_LOAD_CONFIG, NULL)
       └── Marks default context: SKIP config auto-loading
           (OPENSSL_CONF is in env but default context will never read it)

4. Individual test creates ProviderCtx (RAII)
   plugins/ossl_prov/integration-tests/openssl-capi/cpp/utils/provider_ctx.hpp:23-68
   ├── OSSL_LIB_CTX_new()                                                (line 25)
   │   └── Creates a fresh, isolated library context (not the default)
   ├── std::getenv("OPENSSL_CONF")                                       (line 35)
   └── OSSL_LIB_CTX_load_config(libctx_, conf)                           (line 44)
       └── Loads config INTO the isolated context only
           └── activate=1 → OSSL_PROVIDER_load("azihsm") into libctx_
               └── Provider initializes in isolated context
                   └── Provider's OpenSSL API calls use the process-wide libcrypto
                       but the DEFAULT context was never loaded with provider config
                       → no recursive activation, no deadlock

5. Test runs cryptographic operations against ProviderCtx::libctx()
   e.g. EVP_PKEY_CTX_new_from_name(libctx, "EC", "?provider=azihsm")

6. ProviderCtx destructor (~ProviderCtx)                                 (line 70-76)
   └── OSSL_LIB_CTX_free(libctx_)
       └── Unloads provider, releases all resources for that context
```

**Key requirement**: `OPENSSL_INIT_NO_LOAD_CONFIG` must be called before any other OpenSSL API use. This is why it's in `main()` before `InitGoogleTest`. The per-test `OSSL_LIB_CTX` isolation also ensures tests don't interfere with each other.

PR #236, opened 2026-03-04
feat(ossl_prov): add support for importing wrapped keys

walterchris:
This commit moves the duplicated key import logic into shared functions and adds support for importing wrapped keys. Wrapped keys can now also be passed into the azihsm.input_key parameter and it will auto-detect if passed-in key is a wrapped key, or if it's a DER format.

### DIFF
--- a/plugins/ossl_prov/inc/azihsm_ossl_ec.h
+++ b/plugins/ossl_prov/inc/azihsm_ossl_ec.h
@@ -47,9 +47,9 @@ typedef struct
     azihsm_handle session;
     AZIHSM_OSSL_PROV_CTX *provctx;
     bool session_flag;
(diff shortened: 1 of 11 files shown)


plugins/ossl_prov/inc/azihsm_ossl_ec.h
```diff
@@ -49,6 +49,7 @@ typedef struct
     bool session_flag;
     char masked_key_file[4096];
     char input_key_file[4096];
+    char wrapped_key_file[4096];
```
rajesh-gali: we should use a macro to define max file size? I see 4096 magic number in multiple places
```suggestion
    char wrapped_key_file[AZIHSM_MAX_FILE_SIZE];
```

PR #238, opened 2026-03-05
Reject RSA-PSS saltlen=auto as unsupported

### LINKED ISSUE #225: RSA-PSS verification fails when signature was created with auto salt length by OpenSSL
When an RSA-PSS signature is created by the OpenSSL with rsa_pss_saltlen:auto and then verified by our provider (also with auto), verification fails with RsaVerifyError.

jenstopp:
The HSM API requires an explicit salt length and cannot auto-detect it by inspecting a PSS signature. Previously, saltlen=auto was silently accepted and fell back to hash-length behaviour (identical to "digest"), giving misleading results without any error.

Changes:
- Reject "auto" (string) and -2 (integer) in set_ctx_params with PROV_R_INVALID_SALT_LENGTH
- Replace silent AUTO fallback in all four salt-length resolution blocks (one-shot sign/verify, streaming sign/verify init) with explicit errors
- Update AZIHSM_RSA_PSS_SALTLEN_AUTO comment in header to reflect unsupported status
- Update README: replace saltlen:auto verify example with saltlen:digest and remove auto from the PSS options table

Closes #225

### DIFF
--- a/plugins/ossl_prov/README.md
+++ b/plugins/ossl_prov/README.md
@@ -308,7 +308,7 @@ openssl dgst -sha256 ${PROV} \
 
 openssl dgst -sha256 ${PROV} \
     -sigopt rsa_padding_mode:pss \
-    -sigopt rsa_pss_saltlen:auto \
+    -sigopt rsa_pss_saltlen:digest \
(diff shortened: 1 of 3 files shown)


mhatrevi (approved):
:shipit:

PR #306, opened 2026-04-06
[enhance] Enforce streaming context lifecycle - Prevent context reuse

rajesh-gali:


api/lib/src/algo/aes/gcm.rs
```diff
@@ -510,6 +536,11 @@ impl HsmDecryptContext for HsmAesGcmDecryptContext {
         &mut self,
         plaintext: Option<&mut [u8]>,
     ) -> Result<usize, <Self::Algo as HsmDecryptStreamingOp>::Error> {
+        //finish can only be called once successfully, subsequent calls should return error
```
mhatrevi: Replace this with "// Finish can only be called once successfully, subsequent calls should return error"

PR #312, opened 2026-04-09
[Test] add tests to improve rsa key_test  coverage

meilu-git:


api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -574,3 +827,571 @@ fn test_rsa_2048_imported_key_report(session: HsmSession) {
...
+fn test_rsa_functional_4096(session: HsmSession) {
+    run_rsa_functional_test(&session, 4096, 512, 16);
+}
+
+/// Ensure wrong unwrapping key differs for 2048-bit RSA
+#[session_test]
+fn test_rsa_wrong_key_2048(session: HsmSession) {
+    run_rsa_wrong_key_test(&session, 2048, 256, 32);
+}
+
```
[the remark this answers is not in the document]
meilu-git: removed the duplicate

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -387,8 +393,287 @@ fn test_rsa_key_unmask_for_bits(
...
+                "Wrong key must not yield identical key material"
+            );
+
+            HsmKeyManager::delete_key(bad_priv).unwrap();
+        }
+    }
+
+    HsmKeyManager::delete_key(good_priv).unwrap();
+}
+/// Helper to verify repeated unwrap produces usable (not necessarily identical) keys
```
rajesh-gali, line 447: missing new line

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -387,8 +393,287 @@ fn test_rsa_key_unmask_for_bits(
...
+fn run_rsa_repeatability_test(
+    session: &HsmSession,
+    bits: u32,
+    key_size_bytes: usize,
+    salt_len: usize,
+) {
+    use crypto::*;
```
rajesh-gali: can avoid local "use statements?

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -387,8 +393,287 @@ fn test_rsa_key_unmask_for_bits(
...
+}
 
+/// Helper to verify truncated wrapped key fails during unwrap
+fn run_rsa_truncated_ciphertext_test(
+    session: &HsmSession,
+    bits: u32,
+    key_size_bytes: usize,
+    salt_len: usize,
+) {
+    use crypto::*;
```
rajesh-gali: same

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -574,3 +866,466 @@ fn test_rsa_2048_imported_key_report(session: HsmSession) {
     HsmKeyManager::delete_key(priv_key).expect("Failed to delete RSA private key");
     HsmKeyManager::delete_key(pub_key).expect("Failed to delete RSA public key");
 }
+
+/// Ensure unwrap fails when bits do not match actual key size
+#[session_test]
+fn test_unwrap_rsa_wrong_bits_fails(session: HsmSession) {
+    use crypto::*;
```
rajesh-gali: remove use

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -574,3 +866,466 @@ fn test_rsa_2048_imported_key_report(session: HsmSession) {
...
+    assert!(
+        matches!(result, Err(HsmError::InvalidKeyProps)),
+        "Unwrap should fail with InvalidKeyProps for mismatched bits"
+    );
+}
+
+/// Ensure unwrap fails when wrapped key is corrupted
+#[session_test]
+fn test_unwrap_rsa_tampered_ciphertext(session: HsmSession) {
+    use crypto::*;
```
rajesh-gali: same as above

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -574,3 +866,466 @@ fn test_rsa_2048_imported_key_report(session: HsmSession) {
...
+
+    let priv_props = HsmKeyPropsBuilder::default()
+        .class(HsmKeyClass::Private)
+        .key_kind(HsmKeyKind::Rsa)
+        .bits(2048)
+        .can_decrypt(true)
+        .build()
+        .unwrap();
+
+    // ❌ missing can_encrypt
```
rajesh-gali: avoid unicode emoji in the code

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -387,8 +393,287 @@ fn test_rsa_key_unmask_for_bits(
...
+    //  Attempt unwrap with WRONG private key
+    let bad_result = HsmKeyManager::unwrap_key_pair(
+        &mut unwrap_algo,
+        &unwrap_priv_bad,
+        &wrapped,
+        priv_props,
+        pub_props,
+    );
+
+    match bad_result {
```
rajesh-gali: Not sure I understand the logic here, function accepts both Err and Ok,

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -574,3 +866,466 @@ fn test_rsa_2048_imported_key_report(session: HsmSession) {
...
+    let result = HsmKeyManager::unwrap_key_pair(
+        &mut unwrap_algo,
+        &unwrap_priv,
+        &wrapped,
+        priv_props,
+        pub_props,
+    );
+
+    assert!(result.is_err());
+}
```
rajesh-gali, line 1334: missing corrupted unmask blob tests similar to wrap.

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -574,3 +866,466 @@ fn test_rsa_2048_imported_key_report(session: HsmSession) {
...
```
rajesh-gali, line 1292: please add wrong key kind test too

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -387,8 +393,287 @@ fn test_rsa_key_unmask_for_bits(
     HsmKeyManager::delete_key(unmasked_pub_key).expect("Failed to delete unmasked public key");
```
rajesh-gali, line 390: would it be possible to add a test validating round trip with unmask key handle?

like Encrypt with original key and decrypt with unmasked key handle.

rajesh-gali (changes requested):

api/tests/src/algo/rsa/key_tests.rs
meilu-git: fixed

api/tests/src/algo/rsa/key_tests.rs
meilu-git: fixed by moving it to the header of the file and remove local use.

api/tests/src/algo/rsa/key_tests.rs
meilu-git: fixed by moving it to the header of the file and remove local use.

api/tests/src/algo/rsa/key_tests.rs
meilu-git: fixed by moving it to the header of the file and remove local use.

api/tests/src/algo/rsa/key_tests.rs
meilu-git: fixed

api/tests/src/algo/rsa/key_tests.rs
meilu-git: add corruped/truncated/empty unmask blob

api/tests/src/algo/rsa/key_tests.rs
meilu-git: added

api/tests/src/algo/rsa/key_tests.rs
meilu-git: unwrap with a wrong private key is not guaranteed to fail. It may either return Err or succeed with different (invalid) key material. The test ensures that any successful result does not match the original key. clarify this in the code and make it less confusing.

api/tests/src/algo/rsa/key_tests.rs
meilu-git: added test to do the round trip tests ( encrypt with original + decrypt with unmask and encrypt with unmask and decrypt with original)

api/tests/src/algo/rsa/key_tests.rs
meilu-git: resp;ved

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -387,8 +390,398 @@ fn test_rsa_key_unmask_for_bits(
...
+}
+
+/// Verify roundtrip works between original and unmasked key pair
+fn run_rsa_unmask_roundtrip_test(
+    session: &HsmSession,
+    bits: u32,
+    key_size_bytes: usize,
+    salt_len: usize,
+) {
+    use crypto::*;
```
rajesh-gali: still local use

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -387,8 +390,398 @@ fn test_rsa_key_unmask_for_bits(
...
+
+    let mut wrap_algo = HsmRsaAesWrapAlgo::new(HsmHashAlgo::Sha256, salt_len);
+    let wrapped = HsmEncrypter::encrypt_vec(&mut wrap_algo, &unwrap_pub, der)
+        .expect("Failed to wrap RSA key");
+
+    (unwrap_priv, wrapped)
+}
+
+fn get_rsa_masked_blob(session: &HsmSession) -> (Vec<u8>, HsmRsaPrivateKey) {
+    use crypto::*;
```
rajesh-gali: same

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -387,8 +390,398 @@ fn test_rsa_key_unmask_for_bits(
...
+    assert!(
+        matches!(result, Err(HsmError::DdiCmdFailure)),
+        "Truncated ciphertext should fail"
+    );
+}
+
+/// Generates DER-encoded RSA private key of given byte size.
+fn generate_rsa_der(bytes: usize) -> Vec<u8> {
+    use azihsm_crypto::ExportableKey;
+    use azihsm_crypto::KeyGenerationOp;
```
rajesh-gali: local use

api/tests/src/algo/rsa/key_tests.rs
```diff
@@ -387,8 +390,398 @@ fn test_rsa_key_unmask_for_bits(
...
+    if let Ok((bad_priv, _)) = bad_result {
+        let bad_mask = bad_priv.masked_key_vec().unwrap();
+
+        assert_ne!(
+            good_mask, bad_mask,
+            "Wrong key must not yield identical key material"
+        );
+
+        HsmKeyManager::delete_key(bad_priv).unwrap();
+    }
```
rajesh-gali: this will silently pass if bad_result is err ? I think test should check if bad_rsult is error or not. I believe succesfull unwraping with a wrong private key itself is a failure and should never happen

api/tests/src/algo/rsa/key_tests.rs
meilu-git: removed

api/tests/src/algo/rsa/key_tests.rs
meilu-git: removed

api/tests/src/algo/rsa/key_tests.rs
meilu-git: removed

api/tests/src/algo/rsa/key_tests.rs
meilu-git: Based on the discussion, I’m going to remove these tests.
Both unwrap_priv_good and unwrap_priv_bad are generated via get_rsa_unwrapping_key_pair, and in the current test environment the HSM returns the same underlying key, so they are effectively identical.

I explored a few options to obtain distinct keys (e.g., separate sessions via helpers), but those approaches either require session negotiation not exposed in the test API .

Given this, a true “wrong key” scenario cannot be reliably constructed here, so the test does not provide meaningful coverage.

rajesh-gali (approved):

rajesh-gali (approved):

PR #315, opened 2026-04-10
fix: remove sudo as a requirement for NGINX

### LINKED ISSUE #309: Investigate to avoid sudo access to NGINX test
Investigate if it's possible to run all provider tests from local target folder instead of copying to system folders. This could eliminate need for sudo access. 
Or an option in nextest to accept sudo password.

walterchris:


plugins/ossl_prov/integration-tests/nginx/src/nginx_tests.rs
```diff
@@ -106,7 +106,7 @@ mod integration {
             }
         }
 
-        stop_nginx();
+        stop_nginx(&keymat_dir);
```
rajesh-gali: nginx is stopped only if test is successfull, I think nginx needs to be stopped if it has been started or start_nginx fails?

rajesh-gali: we could use drop trait to cleanup nginx ?

rajesh-gali (dismissed):

plugins/ossl_prov/integration-tests/nginx/src/nginx_tests.rs
```diff
@@ -378,91 +386,104 @@ http {{
...
+            let error_log = self.keymat_dir.join("logs").join("error.log");
+            let openssl_conf = self.keymat_dir.join("openssl-provider.cnf");
+            let _ = Command::new("nginx")
+                .args(["-s", "stop"])
+                .arg("-p")
+                .arg(&self.keymat_dir)
+                .arg("-e")
+                .arg(&error_log)
+                .arg("-c")
+                .arg(&nginx_conf)
```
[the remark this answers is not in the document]
walterchris: Dismissed.

rajesh-gali (approved):

PR #346, opened 2026-04-28
Move OpenSSL into setup

### LINKED ISSUE #345: [task] Move openssl installation to setup phase from precheck
Openssl installation is currently installed during precheck. Investigate possibility of moving installation to setup phase instead of precheck, and precheck can report error if OSSL installation is not available.

walterchris:


xtask/src/integration_tests.rs
```diff
@@ -16,70 +15,67 @@ use crate::XtaskCtx;
...
-        let keymat_dir = ctx.root.join("target").join("test-keymat");
-        if keymat_dir.exists() {
-            std::fs::remove_dir_all(&keymat_dir)?;
-            log::trace!(
-                "cleaned previous test key material at {}",
-                keymat_dir.display()
-            );
-        }
+            let keymat_dir = _ctx.root.join("target").join("test-keymat");
+            if keymat_dir.exists() {
```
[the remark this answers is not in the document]
walterchris: clippy is not complaining.

xtask/src/openssl_install.rs
```diff
@@ -109,17 +96,13 @@ pub fn ensure_openssl() -> anyhow::Result<PathBuf> {
     cmd!(sh, "tar xz -C /tmp -f {tarball}").run()?;
```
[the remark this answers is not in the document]
walterchris: ack.

.github/workflows/rust.yml
```diff
@@ -37,18 +33,8 @@ jobs:
...
-    - name: Build OpenSSL 3.0.3 (shared)
-      if: steps.cache-openssl.outputs.cache-hit != 'true'
-      run: |
-        curl -fsSL "https://github.com/openssl/openssl/releases/download/openssl-3.0.3/openssl-3.0.3.tar.gz" \
-          | tar xz -C /tmp
-        cd /tmp/openssl-3.0.3
-        ./Configure --prefix=/opt/openssl-3.0.3 --libdir=lib
-        make -j"$(nproc)"
-        sudo make install_sw
+        path: target/openssl-3.0.3
```
rajesh-gali: I think it would be nice to use a variable/key for target folder path since we are using local target. It would help in future to change the OSSL version, without editing entire workflow
 file.

xtask/src/openssl_install.rs
```diff
@@ -68,35 +50,46 @@ pub fn ensure_openssl() -> anyhow::Result<PathBuf> {
...
 
     // Preflight: check required tools before starting a long build.
     let sh = Shell::new()?;
-    for tool in ["curl", "make", "cc"] {
+    for tool in ["curl", "sha256sum", "make", "cc", "perl"] {
         if cmd!(sh, "which {tool}").quiet().run().is_err() {
             anyhow::bail!(
                 "required tool `{tool}` not found. \
-                 Install build prerequisites: sudo apt-get install build-essential curl"
+                 Install build prerequisites: sudo apt-get install build-essential curl perl"
```
rajesh-gali: missing required tools from the comment. sha256sum, make and cc

.github/workflows/rust.yml
```diff
@@ -221,8 +211,8 @@ jobs:
 
     - name: NGINX integration tests
       env:
-        OPENSSL_BIN: /opt/openssl-3.0.3/bin/openssl
-        OPENSSL_LIB: /opt/openssl-3.0.3/lib
+        OPENSSL_BIN: ${{ github.workspace }}/target/openssl-3.0.3/bin/openssl
+        OPENSSL_LIB: ${{ github.workspace }}/target/openssl-3.0.3/lib
         PROVIDER_PATH: /usr/lib/ossl-modules
```
rajesh-gali: would it be possible to use azihsm_provider.so file from the target folder instead of copying to system path?

walterchris: I think we can safely remove the PROVIDER_PATH and don't need to use the system paths anymore.

The only place we need the system path is in the NGINX test for the `libazihsm_api_native.so`.

xtask/src/install.rs
```diff
@@ -45,30 +45,53 @@ impl Xtask for Install {
...
+        let mut command_args = vec!["--locked".to_string()];
         if self.force {
             command_args.push("--force".to_string());
         }
-
         if self.no_default_features {
             command_args.push("--no-default-features".to_string());
         }
-
         if let Some(features) = self.features {
```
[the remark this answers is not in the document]
walterchris: This is false positive - compiles just fine.

rajesh-gali (approved):

PR #378, opened 2026-05-10
[test]  add RSA nist tests

meilu-git:


api/tests/src/algo/rsa/nist_tests.rs
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    let (priv_key, pub_key) = import_rsa_sign_verify_key(session, vector.priv_der, bits);
+    let hash = hash_test_vector_message(session, vector.hash_algo, vector.msg);
+
+    let mut algo = HsmRsaSignAlgo::with_pkcs1_padding(hsm_hash_from_test(vector.hash_algo));
+
+    let signature = HsmSigner::sign_vec(&mut algo, &priv_key, &hash)
+        .unwrap_or_else(|_| panic!("vector {idx}: PKCS#1 NIST signing failed"));
+
```
rajesh-gali, line 171: test should compare sigature with vector signature, else this test will be same as sign_verify round trip.

api/tests/src/algo/rsa/nist_tests.rs
```diff
@@ -0,0 +1,442 @@
...
+    (signature.len() * 8) as u32
+}
+
+/// Infers RSA key size from ciphertext length.
+fn rsa_bits_from_ciphertext(ciphertext: &[u8]) -> u32 {
+    (ciphertext.len() * 8) as u32
+}
+
+/// Returns whether a vector should be skipped due to known OpenSSL strictness.
+fn should_skip_known_openssl_vector(idx: usize) -> bool {
```
rajesh-gali, line 84: please add more details, why vector 29 is being skipped.

api/tests/src/algo/rsa/common.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use super::*;
+
+/// Generates an RSA key pair configured for RSA-AES wrapping and unwrapping.
+pub fn get_rsa_unwrapping_key_pair(session: &HsmSession) -> (HsmRsaPrivateKey, HsmRsaPublicKey) {
```
rajesh-gali: can we use any existing helper function? instead of creating new one here??

api/tests/src/algo/rsa/key_prop_tests.rs
```diff
@@ -4,7 +4,7 @@
 use azihsm_api::*;
 use azihsm_api_tests_macro::*;
 
-use super::key_tests::*;
+use super::common::*;
```
rajesh-gali: gen_rsa_unwrapping_key_pair seems to be duplicate of get_rsa_unwrapping_key_pair

api/tests/src/algo/rsa/mod.rs
```diff
@@ -7,4 +7,7 @@ mod key_prop_tests;
 mod key_tests;
 mod sign_tests;
 
+mod nist_tests;
 use super::*;
+
+pub mod common;
```
rajesh-gali: don't need pub visibility

api/tests/src/algo/rsa/mod.rs
```diff
@@ -7,4 +7,7 @@ mod key_prop_tests;
 mod key_tests;
 mod sign_tests;
 
+mod nist_tests;
 use super::*;
```
rajesh-gali, line 14: missing mod ordering.

api/tests/src/algo/rsa/common.rs
rajesh-gali: also doesn't have to be pub

crates/crypto/src/rsa/tests/rsa_hash_sign_verify_pkcs1.rs
```diff
@@ -2,6 +2,8 @@
 // Licensed under the MIT License.
 //! Tests for validating RSA signing and verification with PKCS#1 v1.5 padding.
 use super::*;
+use crate::testvectors::rsa::RSA_PKCS1_TEST_VECTORS;
+//use crate::testvectors::rsa::RsaPkcs1TestVector;
```
rajesh-gali: remove commented code

rajesh-gali (changes requested):

api/tests/src/algo/rsa/nist_tests.rs
meilu-git: made the change per comments.

api/tests/src/algo/rsa/common.rs
meilu-git: this is the existing ones, enc_tests use it, as well as sign_tests and hash_sign_test( somehow I didn't add those two to commits) just added.    and changed it to pub (crate)

api/tests/src/algo/rsa/mod.rs
meilu-git: fixed per suggestions

api/tests/src/algo/rsa/mod.rs
meilu-git: fixed per suggestsions

api/tests/src/algo/rsa/key_prop_tests.rs
meilu-git: It is similar, but not an exact duplicate. `get_rsa_unwrapping_key_pair` is a convenience helper that always creates a valid 2048-bit wrapping/unwrapping RSA key pair and panics on failure.

`gen_rsa_unwrapping_key_pair` is lower-level and accepts caller-provided key props, so it is useful for negative key-property tests where we intentionally pass unsupported bits or invalid permissions and assert the returned error.

I will keep gen_rsa_unwrapping_key_pair for key_prop_tests as most negative tests use it.

api/tests/src/algo/rsa/nist_tests.rs
meilu-git: added a bit more details

api/tests/src/algo/rsa/nist_tests.rs
```diff
@@ -0,0 +1,453 @@
...
+    if should_skip_unsupported_bits("PKCS#1", idx, bits) {
+        return;
+    }
+
+    let (_priv_key, pub_key) = import_rsa_sign_verify_key(session, vector.priv_der, bits);
+    let mut algo = HsmRsaHashSignAlgo::with_pkcs1_padding(hsm_hash_from_test(vector.hash_algo));
+
+    let is_valid = HsmVerifier::verify(&mut algo, &pub_key, vector.msg, vector.s)
+        .unwrap_or_else(|_| panic!("vector {idx}: PKCS#1 NIST verify failed"));
+
```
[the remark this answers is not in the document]
meilu-git: fixed per suggestions.

api/tests/src/algo/rsa/nist_tests.rs
```diff
@@ -0,0 +1,453 @@
...
+fn rsa_bits_from_ciphertext(ciphertext: &[u8]) -> u32 {
+    (ciphertext.len() * 8) as u32
+}
+
+/// Returns whether a vector should be skipped due to known OpenSSL strictness.
+///
+/// Vector 29 contains a PKCS#1 v1.5 signature whose encoded DigestInfo does not
+/// satisfy OpenSSL's stricter DER validation rules. The HSM/OpenSSL-backed API
+/// rejects that vector during verification even though the vector is kept in the
+/// shared test-vector set for compatibility with other implementations.
```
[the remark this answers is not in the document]
meilu-git: updated comments to address both PKCS and PSS

api/tests/src/algo/rsa/nist_tests.rs
meilu-git: added more dteails

api/tests/src/algo/rsa/nist_tests.rs
```diff
@@ -0,0 +1,460 @@
...
+
+    let (can_sign, can_verify, can_decrypt, can_encrypt) = match usage {
+        ImportedRsaKeyUsage::SignVerify => (true, true, false, false),
+        ImportedRsaKeyUsage::EncryptDecrypt => (false, false, true, true),
+    };
+
+    let priv_key_props = HsmKeyPropsBuilder::default()
+        .class(HsmKeyClass::Private)
+        .key_kind(HsmKeyKind::Rsa)
+        .bits(bits)
```
rajesh-gali: Vector key should be a session key

api/tests/src/algo/rsa/enc_tests.rs
```diff
@@ -4,35 +4,9 @@
...
-    let (priv_key, pub_key) =
-        HsmKeyManager::generate_key_pair(session, &mut algo, priv_key_props, pub_key_props)
-            .expect("Failed to generate unwrapping key");
-
-    (priv_key, pub_key)
-}
-
 /// Import an external RSA private key DER blob into the HSM by wrapping and unwrapping it.
 fn import_rsa_key(
     session: &HsmSession,
```
rajesh-gali, line 11: hash_sign_tests also has same function, can we reuse?

api/tests/src/algo/rsa/common.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use super::*;
+
+/// Generates an RSA key pair configured for RSA-AES wrapping and unwrapping.
+pub(crate) fn get_rsa_unwrapping_key_pair(
```
rajesh-gali, line 7: similar to gen_rsa_unwrapping_key_pair, check if we can reuse it.

api/tests/src/algo/rsa/nist_tests.rs
```diff
@@ -0,0 +1,460 @@
...
+
+/// Imports RSA private-key DER into HSM RSA key handles through RSA-AES wrap/unwrap.
+fn import_rsa_key_pair(
+    session: &HsmSession,
+    der: &[u8],
+    bits: u32,
+    usage: ImportedRsaKeyUsage,
+) -> (HsmRsaPrivateKey, HsmRsaPublicKey) {
+    let (unwrapping_priv_key, unwrapping_pub_key) = get_rsa_unwrapping_key_pair(session);
+
```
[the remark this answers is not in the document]
meilu-git: I’d prefer to keep this as-is for now. Each vector import currently exercises the full RSA-AES wrap/unwrap import path independently, which keeps the helper simple and avoids passing unwrapping key handles through every vector helper. The imported vector keys are session keys, so cleanup should be tied to the test session. Unless we see actual CI slowdown or handle-table pressure, I think the simpler test structure is preferable here.

api/tests/src/algo/rsa/nist_tests.rs
meilu-git: make it a session key

api/tests/src/algo/rsa/common.rs
meilu-git: get_rsa_unwrapping_key_pair  is in the commons.   however gen_rsa_unwrapping_key_pair is mostly used for negative tests.   I'd keep it as is in prop_key_test

api/tests/src/algo/rsa/enc_tests.rs
meilu-git: made a common helper in common.rs.  each tests will call the helper with the right parameter ( sign/verify, encrypt/decrypt,  session is true) etc

api/tests/src/algo/rsa/common.rs
```diff
@@ -0,0 +1,91 @@
...
+/// Common helper that imports RSA private-key DER into HSM RSA key handles
+/// through RSA-AES wrap/unwrap.
+pub(crate) fn try_import_rsa_key_pair(
+    session: &HsmSession,
+    der: &[u8],
+    bits: u32,
+    usage: ImportedRsaKeyUsage,
+    is_session: bool,
+) -> Result<(HsmRsaPrivateKey, HsmRsaPublicKey), HsmError> {
+    let (unwrapping_priv_key, unwrapping_pub_key) = get_rsa_unwrapping_key_pair(session);
```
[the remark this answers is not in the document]
meilu-git: I’d prefer not to add that refactor in this PR. The helper intentionally exercises the full RSA-AES import path independently for each vector, and the imported vector keys are now session keys. Reusing one unwrapping key pair could improve performance, but it would add extra plumbing through all PKCS#1/PSS/OAEP helpers. Unless we see actual CI slowdown or handle pressure, I think the simpler test structure is preferable.

api/tests/src/algo/rsa/nist_tests.rs
```diff
@@ -0,0 +1,413 @@
...
+    }
+}
+
+/// Returns whether the RSA key size is supported by the HSM API tests.
+fn is_supported_rsa_bits(bits: u32) -> bool {
+    matches!(bits, 2048 | 3072 | 4096)
+}
+
+/// Imports RSA private-key DER into HSM RSA key handles through RSA-AES wrap/unwrap.
+/// Imports RSA private-key DER into HSM RSA session key handles through RSA-AES wrap/unwrap.
```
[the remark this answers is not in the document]
meilu-git: removed dup

api/tests/src/algo/rsa/nist_tests.rs
```diff
@@ -0,0 +1,413 @@
...
+        println!("Skipping {kind} vector {idx}: unsupported RSA key size {bits}");
+        true
+    }
+}
+
+/// Hashes a message using the hash algorithm specified by the RSA test vector.
+fn hash_test_vector_message(session: &HsmSession, hash_algo: TestHashAlgo, msg: &[u8]) -> Vec<u8> {
+    let mut hash_algo = hsm_hash_from_test(hash_algo);
+
+    HsmHasher::hash_vec(session, &mut hash_algo, msg)
```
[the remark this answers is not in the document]
meilu-git: fixed per suggestions

api/tests/src/algo/rsa/nist_tests.rs
```diff
@@ -0,0 +1,413 @@
...
+/// Imports RSA private-key DER into HSM RSA key handles through RSA-AES wrap/unwrap.
+/// Imports RSA private-key DER into HSM RSA session key handles through RSA-AES wrap/unwrap.
+fn import_rsa_key_pair(
+    session: &HsmSession,
+    der: &[u8],
+    bits: u32,
+    usage: ImportedRsaKeyUsage,
+) -> (HsmRsaPrivateKey, HsmRsaPublicKey) {
+    try_import_rsa_key_pair(session, der, bits, usage, true)
+        .expect("Failed to RSA-AES unwrap RSA session key pair")
```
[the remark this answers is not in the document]
meilu-git: fixed per suggestions.

rajesh-gali (approved):

PR #384, opened 2026-05-12
Do not override the normal algorithm for finding config.toml files

netsweng:
Do not override the normal algorithm for finding config.toml files as the 2 choices encoded here do not cover all situations. Instead, leave the default to let cargo do it's normal thing, but add a --config option so pipelines can pass something in for unusual cases.


xtask/src/precheck.rs
```diff
@@ -119,21 +120,9 @@ impl Xtask for Precheck {
...
-                    anyhow::bail!("Could not find .cargo directory at {}", config_path);
-                }
-            }
-
-            config_path.push_str("/config.toml");
-
             Setup {
                 force: false,
-                config: Some(config_path),
+                config: self.config,
```
jaygmsft, line 126: will users need to pass this config when running precheck either when enlisted with azihsm-sdk or its other counterpart location?

netsweng: mostly in the other location. Default in azihsm-sdk seem to do the right thing already.

xtask/src/precheck.rs
```diff
@@ -95,14 +94,16 @@ pub struct Precheck {
     /// The nextest profile to use
     #[clap(long)]
     profile: Option<String>,
+    /// Pass through to `cargo install --config`; accepts either `KEY=VALUE`
+    /// or a path to a Cargo `config.toml` file.
```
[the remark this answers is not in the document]
netsweng: change commited.

xtask/src/precheck.rs
jaygmsft: We need to ensure that cargo xtask precheck can work on any location that we regularly use for development without having to pass additional arguments manually.

jaygmsft: pipelines can pass overrides and that is ok, user shouldnt have to pass it for local runs on devbox

netsweng: config.toml is in the expected location withing the azihsm-sdk tree. By removing the hardcoded paths here, the default behavior of walking up the tree as many levels as is needed should once again be in effect.

jaygmsft (approved):

PR #390, opened 2026-05-20
refactor(fw): make MBOR decoder and request types DMA-aware

jaygmsft:
The MborDecoder.buffer field and decoded byte-slice fields in MBOR request types now use &DmaBuf instead of &[u8]. Since the firmware DDI request buffer is always DMA memory, this propagates the DMA type all the way to PAL crypto calls, eliminating staging copies in handlers that previously had to copy borrowed wire data into DMA buffers before invoking crypto primitives.

Changes:
- fw/pal/traits: add PartialEq / Eq / PartialEq<[u8]> / PartialEq<[u8;N]>
  impls on DmaBuf so test assertions and equality checks work directly
  on &DmaBuf without explicit dereference.
- fw/core/ddi/mbor/codec: MborDecoder now stores &'a DmaBuf and returns &DmaBuf from bytes(), decode_byte_slice(), and decode_byte_slice_exact(). try_into() coercions go through an explicit &[u8] binding so the array conversion still resolves.
- fw/core/ddi/mbor/api: DdiDecoder::new takes &DmaBuf.
- fw/core/ddi/mbor/types: all Req struct byte-slice fields and shared types (DdiPublicKey.raw, DdiKeyProperties.key_label, DdiTargetKeyProperties.key_label) now use &'a DmaBuf. Resp fields stay &[u8] since they are constructed from mixed sources (stack arrays, byte literals, DMA frame slots).
- fw/core/ddi/mbor/types/{change_pin, open_session}: fix max_len -> len on fixed-size AES-encrypted fields (encrypted_id, encrypted_pin, iv, encrypted_seed) that were incorrectly marked variable-length.
- fw/core/lib/src/ddi/sha_digest.rs: drop the staging copy and pass body.msg directly to pal.hash().
- Update unit and cross-compat tests to use a dma() helper that wraps DmaBuf::from_raw for test slices.

Host-side MBOR (azihsm_ddi_mbor) is unchanged; this only affects the firmware-side codec.

Tests:
- fw mbor + types + api + core + pal_traits unit tests: 61/61 pass.
- emu smoke (azihsm_ddi --features emu, smoke filter): 10/10 pass.

### DIFF
--- a/fw/pal/traits/src/alloc.rs
+++ b/fw/pal/traits/src/alloc.rs
@@ -167,26 +167,6 @@ impl core::fmt::Debug for DmaBuf {
     }
 }
 
-impl PartialEq for DmaBuf {
-    fn eq(&self, other: &Self) -> bool {
-        self.inner == other.inner
-    }
-}
-
-impl PartialEq<[u8]> for DmaBuf {
-    fn eq(&self, other: &[u8]) -> bool {
-        &self.inner == other
-    }
-}
-
-impl<const N: usize> PartialEq<[u8; N]> for DmaBuf {
-    fn eq(&self, other: &[u8; N]) -> bool {
-        &self.inner == other
-    }
-}
-
-impl Eq for DmaBuf {}
-
 impl core::ops::Index<usize> for DmaBuf {
     type Output = u8;
     #[inline(always)]
--- a/fw/pal/traits/src/alloc.rs
+++ b/fw/pal/traits/src/alloc.rs
@@ -167,6 +167,26 @@ impl core::fmt::Debug for DmaBuf {
     }
 }
 
+impl PartialEq for DmaBuf {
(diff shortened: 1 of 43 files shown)


fw/pal/traits/src/alloc.rs
```diff
@@ -167,6 +167,26 @@ impl core::fmt::Debug for DmaBuf {
...
+    }
+}
+
+impl<const N: usize> PartialEq<[u8; N]> for DmaBuf {
+    fn eq(&self, other: &[u8; N]) -> bool {
+        &self.inner == other
+    }
+}
+
+impl Eq for DmaBuf {}
```
vsonims: can we do explicit comparisions where necessary. Misuse hear leads to perf degradation. It is just a variant of no mem copy.

vsonims (approved):
Please address the feedback and check in

PR #407, opened 2026-05-28
Resiliency fault injection tests fix

mhatrevi:


api/tests/src/utils/partition.rs
```diff
@@ -197,26 +197,42 @@ pub(crate) fn init_with_mobk_fallback(
...
+/// partially-written cache file.
 fn mobk_cache_file_path(_part_path: &str) -> std::path::PathBuf {
     match std::env::var("AZIHSM_MOBK_PATH") {
         Ok(p) if !p.is_empty() => std::path::PathBuf::from(p),
-        _ => std::env::temp_dir().join("mobk.bin"),
+        _ => default_mobk_cache_file_path().clone(),
     }
 }
 
+fn default_mobk_cache_file_path() -> &'static std::path::PathBuf {
```
rajesh-gali, line 218: [Nit] Missing doc comments

api/tests/src/utils/partition.rs
```diff
@@ -197,26 +197,42 @@ pub(crate) fn init_with_mobk_fallback(
...
+    static DEFAULT_MOBK_CACHE_FILE: std::sync::OnceLock<std::path::PathBuf> =
+        std::sync::OnceLock::new();
+
+    DEFAULT_MOBK_CACHE_FILE.get_or_init(|| {
```
rajesh-gali: would it be possible to simplify the logic using LazyLock, which removes the need for get_or_init?

mhatrevi: Updated.

api/tests/src/utils/partition.rs
mhatrevi: Added.

rajesh-gali (approved):

PR #408, opened 2026-05-28
[fix] Return a larger buffer hint for cert chain

mhatrevi:
Return a larger buffer hint for the manufacturer certificate chain property so C API callers that follow the size-query/fetch pattern have room for modest cert-chain growth during reset.

Successful fetches still report the actual bytes copied, and other partition properties continue to use their exact required sizes.

### DIFF
--- a/api/native/src/partition_props.rs
+++ b/api/native/src/partition_props.rs
@@ -180,7 +180,16 @@ fn get_partition_prop(
         }
         AzihsmPartPropId::ManufacturerCertChain => {
             let cert_chain = AzihsmStr::from_string(&partition.cert_chain(0)?);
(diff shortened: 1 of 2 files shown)


api/native/src/partition_props.rs
```diff
@@ -235,6 +258,12 @@ fn copy_to_part_prop(part_prop: &mut AzihsmPartProp, bytes: &[u8]) -> Result<(),
     Ok(())
 }
 
+fn cert_chain_buffer_len_hint(actual_len: usize) -> usize {
```
rajesh-gali, line 268: [Nit] Missing doc comments.

mhatrevi: Added; Also added a test.

rajesh-gali (approved):

PR #409, opened 2026-05-28
Delete mock tests folder before running code coverage

rajesh-gali:


.github/workflows/rust.yml
```diff
@@ -66,6 +66,10 @@ jobs:
       id: test-tbor-emu
       run: cargo xtask precheck --nextest --package azihsm_ddi_tbor_types --features emu --profile ci-tbor-emu
       continue-on-error: true
+    - name: Free target/ before coverage build
+      # Freeing target/ directory to ensure disk space for coverage report generation
+      run: |
+        rm -rf target/debug target/tmp
```
zimmy87: is there any reason why we don't wanna use `cargo xtask clean` here?

.github/workflows/rust.yml
```diff
@@ -66,6 +66,10 @@ jobs:
       id: test-tbor-emu
       run: cargo xtask precheck --nextest --package azihsm_ddi_tbor_types --features emu --profile ci-tbor-emu
       continue-on-error: true
+    - name: Free target/ before coverage build
+      # Freeing target/ directory to ensure disk space for coverage report generation
+      run: |
```
zimmy87: why do we need to use a multiline command here? Can't we just put the next line here?

zimmy87 (changes requested):

.github/workflows/rust.yml
zimmy87: slight correction: `cargo clean` might be ideal here for the time being: the nextest xtask isn't pointing to the xtask target directory like the build xtask, so using the clean xtask will most likely do nothing right now

.github/workflows/rust.yml
rajesh-gali: fixed

.github/workflows/rust.yml
rajesh-gali: cargo clean would also remove report files right?

zimmy87: good point: I forgot about the test report files in target/nextest: we need to keep these for the test summary. Resolving as by design.

zimmy87 (approved):
lgtm

.github/workflows/rust.yml
```diff
@@ -66,6 +66,9 @@ jobs:
       id: test-tbor-emu
       run: cargo xtask precheck --nextest --package azihsm_ddi_tbor_types --features emu --profile ci-tbor-emu
       continue-on-error: true
+    - name: Free target/ before coverage build
+      # Freeing target/ directory to ensure disk space for coverage report generation
+      run: rm -rf target/debug target/tmp
```
jaygmsft, line 71: wont it increase the build times?

PR #417, opened 2026-06-02
feat(fw): implement DDI AesGenerateKey + AesEncryptDecrypt (CBC) hand…

jaygmsft:
AesGenerateKey generates a fresh random AES-128 / 192 / 256 key in the partition vault (optionally session-scoped so it is torn down by CloseSession) and returns the assigned key_id plus a placeholder masked_key envelope.  Bulk AES variants (XTS / GCM) are out of scope and rejected with InvalidArg.

AesEncryptDecrypt performs AES-CBC encrypt or decrypt on a host-supplied message buffer using a vault-resident AES key and a host-supplied 16-byte IV.  The transformed message plus the updated chaining IV are returned so the host can chain subsequent CBC blocks.

Handler details:

- AesGenerateKey:
  - Validates the key_size is non-bulk (Aes128/192/256).
  - Translates DdiTargetKeyMetadata flags into vault attrs; only the EncryptDecrypt usage is valid for AES, anything else returns InvalidPermissions.
  - Session-only keys cannot carry an explicit key_tag (matches test_aes_generate_session_only_key_with_key_tag).
  - Fills a scratch DMA buffer with rng_fill_bytes, then vault_key_create with the appropriate Aes{128,192,256} kind and session binding.
  - bulk_key_id is always None for non-bulk requests.
- AesEncryptDecrypt:
  - Validates the op enum (Encrypt / Decrypt).
  - Validates IV is exactly one AES block, message is non-empty, block-aligned, and within the 1024 B wire max.
  - Looks up the vault key and rejects non-AES kinds with InvalidKeyType.
  - Op-aware permission check: Encrypt requires the encrypt attribute bit, Decrypt requires the decrypt bit.
  - Runs aes_cbc_enc_dec_in_place directly inside the response msg / iv slots — no scratch buffers — and emits the updated chaining IV so the host can stitch multi-block sequences.

New type surface:

- fw-side DdiTargetKeyMetadata bit accessors (session / sign / verify / encrypt / decrypt / derive / unwrap), mirroring the host-side helpers so handlers can read the requested key-usage flags without poking the raw 16-byte blob.

Adds 4 smoke tests (generate happy + generate without session; cbc encrypt+decrypt roundtrip + cbc without session + cbc unknown-key) and brings 24 of the 25 emu integration tests in the aes_generate + aes_cbc modules to passing — the lone failure (test_aes_cbc_encrypt_decrypt_incorrect_key_type) depends on the EccGenerateKeyPair handler, which is in a separate stash branch.

### DIFF
--- a/fw/core/lib/src/ddi/mbor/aes_generate_key.rs
+++ b/fw/core/lib/src/ddi/mbor/aes_generate_key.rs
@@ -1,0 +1,138 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! DDI AesGenerateKey command handler.
+//!
+//! Within an open session, generate a fresh random AES key (128 /
+//! 192 / 256 bits), persist it in the partition vault — optionally
+//! session-scoped so it is torn down by
+//! [`CloseSession`](super::close_session) — and return the assigned
+//! `key_id` plus an (empty placeholder) masked-key envelope that the
+//! host may re-import on a future session.
+//!
+//! Scope: non-bulk AES key kinds only.  XTS / GCM bulk variants are
+//! rejected with `InvalidArg`.
+
+use azihsm_fw_ddi_mbor_types::aes_generate_key::DdiAesGenerateKeyReq;
+use azihsm_fw_ddi_mbor_types::aes_generate_key::DdiAesGenerateKeyResp;
+use azihsm_fw_ddi_mbor_types::DdiAesKeySize;
+
+use super::*;
+
+/// Handle `DdiAesGenerateKeyCmd`.
+///
+/// No `partition_lock` is needed: this handler does not perform any
+/// multi-step read-then-mutate against partition state.  Its single
+/// state mutation — `vault_key_create` — is sync and atomic.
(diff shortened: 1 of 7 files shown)


fw/core/lib/src/ddi/mbor/aes_generate_key.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    // Session-only keys are anonymous — disallow a host-supplied
+    // `key_tag` because the key cannot be looked up across sessions.
+    if attrs.session() && body.key_tag.is_some() {
+        return Err(HsmError::InvalidArg);
+    }
+
+    // Generate the random key into a scratch buffer.
+    let key_buf = pal.dma_alloc(io, key_len)?;
+    pal.rng_fill_bytes(io, key_buf)?;
```
vsonims: We should have aes_gen_key api in PAL

vsonims (changes requested):

fw/core/lib/src/ddi/mbor/aes_generate_key.rs
jaygmsft: done

vsonims (approved):

PR #421, opened 2026-06-02
feat(fw): implement DDI EccGenerateKeyPair + EccSign handlers

jaygmsft:
Add the firmware-side handlers for the two ECC DDI commands, the PAL trait surface that supports them, and matching std-PAL impls.

* DDI handlers:
  * `ecc_generate_key_pair`: validates session + curve, builds vault attrs from `key_metadata` (with `wrap` folded into usage_count so a sign+verify+wrap request fails fast instead of silently being treated as plain sign+verify), generates the keypair via the new PAL query-alloc-use flow inside a single scoped allocator, stores the private key in the partition vault (session-scoped iff requested), and emits the wire-format public key in the response.
  * `ecc_sign`: looks up a vault ECC key, validates the `sign` usage bit, and produces a raw `r || s` signature into the response slot. The handler now hands the wire-LE digest sub-slice directly to the PAL — endianness conversion lives entirely in the std PAL.

* PAL trait surface (`HsmEcc::ecc_gen_keypair`): switched to the query-alloc-use shape — takes a scoped allocator + an `Option<(&mut DmaBuf, &mut DmaBuf)>` for the two output buffers, returns `(usize, usize)` (query: per-curve upper bounds; use: actual bytes written, always `\u{2264}` the query bounds).  For real-HW PALs query == use; for std PAL DER private keys can be shorter than the max.  `ecc_sign` doc pins the wire-LE digest contract.

* std PAL (`fw/plat/std/pal/src/ecc.rs`):
  * `ecc_gen_keypair` implements the new shape: query returns `(priv_key_der_max, wire_pub_key_len)`; use allocates a contiguous `priv_max || pub_max` scratch (mirroring what real PKA emits as a single block), writes PKCS#8 DER + wire-LE coordinates, then copies into the caller's two output slots and returns the actual DER length.
  * `ecc_sign` now reverses the wire-LE digest to OpenSSL-BE internally so the trait stays wire-format consistent with `pub_key` / `signature`.

* Misc:
  * Tests: 6 new smoke tests (ecc_generate / ecc_sign).
  * All 30 emu smoke + 27/28 ECC-targeted tests pass; only the pre-existing unimplemented `attest_ecc_signing_key` still fails.
  * Mock full suite: 580/580.

### DIFF
--- a/fw/core/lib/src/ddi/mbor/ecc_sign.rs
+++ b/fw/core/lib/src/ddi/mbor/ecc_sign.rs
@@ -10,7 +10,6 @@
 
 use azihsm_fw_ddi_mbor_types::ecc_sign::DdiEccSignReq;
 use azihsm_fw_ddi_mbor_types::ecc_sign::DdiEccSignResp;
-use azihsm_fw_ddi_mbor_types::DdiHashAlgorithm;
 
 use super::*;
 
@@ -34,7 +33,7 @@ pub(crate) async fn ecc_sign<'p, P: HsmPal>(
     // responsible for any endianness flip its underlying primitive
     // requires (e.g. std PAL reverses to BE for OpenSSL; real-HW
     // PALs pass through to the PKA engine).
-    let pal_algo = ddi_to_pal_hash(body.digest_algo)?;
+    let pal_algo = super::from_ddi::hash(body.digest_algo)?;
     let real_digest_len = pal_algo.digest_len();
     if body.digest.len() < real_digest_len {
         return Err(HsmError::InvalidArg);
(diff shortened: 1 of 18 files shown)


fw/plat/std/pal/src/ecc.rs
```diff
@@ -51,56 +72,144 @@ fn to_ecc_curve(curve: HsmEccCurve) -> EccCurve {
...
-        // Export public key as raw coordinates (x ∥ y).
-        let pub_key = &mut rest[..coord_len];
-        let half = coord_len / 2;
-        let (x_buf, y_buf) = pub_key.split_at_mut(half);
-        pubk.coord(Some((&mut x_buf[..], &mut y_buf[..])))
+        // Export raw BE coordinates into stack scratch, then reverse
+        // each into the wire-format LE slot (leaving any P-521
+        // trailing pad bytes zero).
+        let coord_len = curve.priv_key_len();
+        let mut x_be = [0u8; 66];
```
vsonims: This will get captured in task storage

fw/plat/std/pal/src/ecc.rs
```diff
@@ -51,56 +72,144 @@ fn to_ecc_curve(curve: HsmEccCurve) -> EccCurve {
...
+        let mut y_be = [0u8; 66];
+        pubk.coord(Some((&mut x_be[..coord_len], &mut y_be[..coord_len])))
             .map_err(|_| HsmError::EccToDerError)?;
 
-        Ok((&*priv_key, &*pub_key))
+        scratch_pub.fill(0);
```
vsonims: There is driver for ecc for std_pal. Should it return LE bytes always. This way the code ad this layer is simplified.

fw/core/lib/src/ddi/mbor/ecc_sign.rs
```diff
@@ -0,0 +1,97 @@
...
+        frame.signature,
+    )
+    .await?;
+
+    Ok(resp)
+}
+
+/// Map the DDI digest enum to its PAL counterpart.
+fn ddi_to_pal_hash(algo: DdiHashAlgorithm) -> HsmResult<HsmHashAlgo> {
+    match algo {
```
vsonims: We might need these conversions for ECDH, HMAC & Hashing. Should we move them to a common file?

vsonims (changes requested):

vsonims (approved):

PR #422, opened 2026-06-03
feat(fw): implement DDI GetUnwrappingKey + RsaUnwrap handlers

jaygmsft:
Add the firmware-side handlers for the two key-wrap DDI commands, the PAL trait surface that supports them, and matching std-PAL impls.

* DDI handlers:
  * `get_unwrapping_key`: returns the partition's RSA-2048 unwrap key (vault id, wire-format pub key, empty masked_key placeholder). Lazy generation under the partition lock so concurrent first-call races never produce two competing keys; cached on partition state across sessions, regenerated on partition reset.
  * `rsa_unwrap`: takes an `OAEP(KEK) || AES-KWP(payload)` wrapped blob, RSA-OAEP-decrypts the KEK with the partition's unwrap key, AES-KWP-unwraps the payload with the KEK, and imports the recovered key into the vault.  Supports AES (16/24/32 byte) and RSA (2k/3k/4k, CRT and non-CRT) imports; ECC import returns `UnsupportedCmd`.  Spec error codes (`RsaUnwrapOaepDecodeFailed`, `RsaUnwrapAesUnwrapFailed`, `RsaUnwrapInvalidRequest`, `RsaUnwrapInvalidKek`) are mapped from the underlying crypto failures.

* PAL trait additions (`HsmRsa`, `HsmPart`):
  * `rsa_gen_keypair` refactored to the query-alloc-use shape (typo `ras_` \u2192 `rsa_`, takes scoped allocator, returns `(priv_actual, pub_actual)`) — the only existing caller path (`get_unwrapping_key`) uses the new shape from day one.
  * `rsa_priv_key_size(io, key)` introspects an imported priv key.
  * `rsa_priv_pub_key(io, key, pub_out)` query-alloc-use extractor for the wire-format pub key from a serialized priv (lets the RSA-import path populate the response's optional `pub_key` so callers can verify the import bytewise without an extra DDI round-trip).
  * `part_unwrapping_pub_key` getter / setter for the cached wire-format unwrap pub key (mirrors the existing `part_establish_cred_pub_key` pattern).
  * `HsmRsaKey` gains `pub_wire_len`, `pub_exp_len`, and `priv_key_der_max` helpers.

* std PAL:
  * `rsa_gen_keypair` emits wire-LE `n || e` with deterministic fixed-width padding (256 + 4 = 260 bytes for RSA-2048).
  * `rsa_oaep_decrypt` implemented \u2014 LE\u2192BE flip on the ciphertext (host wraps under PKA-native LE), OpenSSL OAEP decrypt into a modulus-sized scratch, copy actual plaintext into caller's slot.
  * `aes_kwp_unwrap` implemented (delegates to `AesKeyWrapPadAlgo::decrypt`).
  * `rsa_priv_key_size` / `rsa_priv_pub_key` implemented (parse PKCS#8 DER, read modulus / extract n,e in wire format).
  * Partition state gains `unwrapping_pub_key: [u8; 260]` plus the cleared-on-reset hookup in `clear_enabled_state`.

* `build_attrs_for_aes` now `pub(super)` with a `local: bool` parameter so both `aes_generate_key` (true) and the AES unwrap path (false) can share the usage-bit validation logic.

* Tests: 3 smoke tests for GetUnwrappingKey (happy path, stability across calls, no-session) and 4 smoke tests for RsaUnwrap (AES-256 with local-encrypt-vs-device-decrypt cross-check; RSA-3K both CRT and non-CRT with returned-pub-key vs locally-derived-pub-key bytewise comparison; no-session).  580+ DDI mock tests still pass; emu smoke set (33) all green.


fw/core/lib/src/ddi/mbor/get_unwrapping_key.rs
```diff
@@ -0,0 +1,145 @@
...
+        )?;
+        let layout = DdiGetUnwrappingKeyResp::reserve(
+            &mut encoder,
+            key_id,
+            DdiPublicKeyFrameParams {
+                raw_len: pub_wire_len,
+                key_kind: DdiKeyType::Rsa2kPublic,
+            },
+            0, /* masked_key length — empty placeholder until masking lands */
+        )?;
```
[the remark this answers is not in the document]
jaygmsft: will be done when we add masked key support

fw/core/lib/src/ddi/mbor/rsa_unwrap.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    // Step 1: RSA-OAEP-decrypt the KEK using the unwrap private key.
+    // The output buffer must be at least `max_oaep_message` per the
+    // PAL trait contract (real-HW PALs may decrypt directly into the
+    // caller slot via DMA, so undersizing it would either be
+    // rejected up-front or overflow into adjacent memory).  The
+    // recovered KEK is itself only 16 / 24 / 32 bytes — any other
+    // length is a host contract violation (or corrupted blob).
+    let label_empty = pal.dma_alloc(io, 0)?;
+    let kek_buf = pal.dma_alloc(io, rsa_key_size.max_oaep_message(hash_algo))?;
```
vsonims: can we decrypt in place instead of doing an alloc?

fw/core/lib/src/ddi/mbor/rsa_unwrap.rs
```diff
@@ -0,0 +1,153 @@
...
+                oaep_ct,
+                label_empty,
+                &mut *kek_buf,
+                a,
+            )
+            .await
+        })
+        .await
+        .map_err(|_| HsmError::RsaUnwrapOaepDecodeFailed)?;
+    if !matches!(kek_len, 16 | 24 | 32) {
```
vsonims: we should limit kek  to 32 only

fw/core/lib/src/ddi/mbor/rsa_unwrap/import_aes.rs
```diff
@@ -0,0 +1,59 @@
...
+    key_bytes: &DmaBuf,
+) -> HsmResult<(<P as HsmVault>::KeyGuard<'p>, DdiRsaUnwrapResp<'p>)> {
+    let (vault_kind, key_kind) = match key_bytes.len() {
+        16 => (HsmVaultKeyKind::Aes128, DdiKeyType::Aes128),
+        24 => (HsmVaultKeyKind::Aes192, DdiKeyType::Aes192),
+        32 => (HsmVaultKeyKind::Aes256, DdiKeyType::Aes256),
+        _ => return Err(HsmError::InvalidArg),
+    };
+
+    let attrs = super::super::key_attrs::prepare_aes(
```
vsonims: this is interesting

fw/plat/std/pal/src/aes.rs
```diff
@@ -245,11 +245,24 @@ impl HsmAes for StdHsmPal {
...
         _io: &impl HsmIo,
-        _key: &DmaBuf,
-        _input: &DmaBuf,
-        _output: &mut DmaBuf,
+        key: &DmaBuf,
+        input: &DmaBuf,
+        output: &mut DmaBuf,
     ) -> HsmResult<usize> {
-        todo!()
+        use azihsm_crypto::AesKey;
```
vsonims: should this be at top of the file.

fw/plat/std/pal/src/aes.rs
```diff
@@ -245,11 +245,24 @@ impl HsmAes for StdHsmPal {
...
+        use azihsm_crypto::AesKeyWrapPadAlgo;
+        use azihsm_crypto::Decrypter;
+        use azihsm_crypto::ImportableKey;
+
```
vsonims: should this be in aes driver?

fw/plat/std/pal/src/ecc.rs
```diff
@@ -51,6 +53,15 @@ fn to_ecc_curve(curve: HsmEccCurve) -> EccCurve {
     }
 }
 
+/// Reverse of [`to_ecc_curve`].
```
vsonims, line 56: This belongs in the driver.

fw/plat/std/pal/src/ecc.rs
```diff
@@ -185,4 +196,61 @@ impl HsmEcc for StdHsmPal {
...
+            return Ok(hsm_len);
+        };
+
+        if out.len() < hsm_len {
+            return Err(HsmError::InvalidArg);
+        }
+        pk.to_hsm_bytes(&mut out[..hsm_len])
+            .map_err(|_| HsmError::EccExportError)?;
+        Ok(hsm_len)
+    }
```
vsonims, line 255: we should not be doing der decode in pal

fw/plat/std/pal/src/rsa.rs
```diff
@@ -22,32 +22,122 @@ fn key_size_bits(key_size: HsmRsaKey) -> usize {
...
+    if pub_out.len() != total {
+        return Err(HsmError::InvalidArg);
+    }
+
+    // Extract the BE components into stack scratch (any modulus up
+    // to RSA-4096 fits in 512 bytes), then reverse each into the
+    // fixed-width wire-LE slot.  Right-align (leading-zero pad)
+    // before reversal so the on-wire layout is independent of how
+    // many leading zero bytes OpenSSL stripped.
+    let mut n_be = [0u8; 512];
```
vsonims, line 56: this will blow up async state

vsonims (changes requested):

PR #415, opened 2026-05-31
ci: add provider integration test matrix

walterchris:
Runs the OpenSSL provider integration suites (cli + capi + nginx) across five OpenSSL 3.0.x distributions in parallel. The nginx suite runs only where OpenSSL carries the complete STORE-dispatch fix (openssl#18262) — the 3.0.11/3.0.13 cells — and is skipped on 3.0.2 and the RHEL-family 3.0.7 cells.

PR #442, opened 2026-06-08
Openssl 3.5 fix

jenstopp:
Fix two OpenSSL-3.5 problems blocking the Ubuntu 26.04 / OpenSSL 3.5.5
integration cell added in #415. No OpenSSL code/version or provider C-code
changes. **Building on #415 — merge first or apply on top.**

Result: 26.04 / 3.5.5 cell green end-to-end (CLI 27/27, CAPI, nginx); all 3.0.x
cells stay green.

## 1. CLI + CAPI — mock-SDK crypto re-entered the provider on 3.5

`crates/crypto` (the simulated device's software crypto, used in `--features
mock`) computes SHA/HMAC/AES/RSA/ECDH/HKDF via libcrypto in the **process-default
`OSSL_LIB_CTX`**. On 3.5 a bare (no-propquery) fetch in a context with azihsm
loaded resolves to **azihsm** (on 3.0.x it resolved to default). Since the
provider opens its HSM session lazily (#403), azihsm is already active when the
mock device runs its own crypto → those fetches re-enter azihsm and every op
fails at init (`gen_init: initialization error`).

**Fix:** new `crates/crypto/src/libctx.rs` — a process-private `OSSL_LIB_CTX`
holding only the `default` provider; every `*_ossl` backend fetches from it
(`Md::fetch`/`Cipher::fetch`/`kdf::hkdf` with `Some(crypto_libctx())`; HMAC and
the EVP_PKEY ops via `EVP_MAC_fetch` / `EVP_PKEY_CTX_new_from_pkey(libctx,…)`,
hand-declared as they're unbound in openssl-sys 0.9.x). Provider C code, lazy
open, and the #403 re-entry guard are untouched.

**No 3.0 regression:** a default-only context resolves to default on *every*
OpenSSL version — it no longer depends on the 3.5-vs-3.0 bare-fetch tiebreak.

**Three orthogonal mechanisms** — this PR adds the middle one on top of #403's two:

- **Lazy open** (#403) = *WHEN* the session opens. Keeps SDK crypto out of
  libcrypto's own init; reverting it to eager fixes 3.5 but regresses 3.0
  (`RAND_set_DRBG_type: odd number of digits`), so it's kept.
- **Private libctx** (this PR) = *WHICH* provider serves the SDK's fetches.
- **Re-entry guard** (#403) = deadlock safety on the open path. Its main 3.5
  trigger is now gone, but it still backstops the provider's own `RAND_bytes`
  DRBG primer during open, so it's retained.

## 2. nginx — three independent blockers on 3.5.5

1. **QUIC HKDF hijacked.** At `nginx -t`, the HTTP/3 module derives the QUIC
   av_token key (`ngx_hkdf_extract` → `EVP_PKEY_CTX_new_id(EVP_PKEY_HKDF)`,
   extract-only) even with no QUIC listener. With azihsm loaded it resolves to
   azihsm's HKDF (extract-and-expand only) → `Provider routines::failed to get
   parameter`. **Fix:** `default_properties = ?provider=default` in nginx's
   OpenSSL config — nginx's own crypto prefers default; the HSM cert key still
   loads via its explicit `store:azihsm://` URI and the optional `?` lets
   key-bound ops fall back to azihsm. (Standard "key-only provider alongside
   default" pattern.)
2. **`store:` key unsupported.** The cell's distro nginx (1.28.x) predates the
   `store:` scheme (nginx mainline 1.29.0). **Fix:** pin nginx 1.29.8 from
   nginx.org for the 26.04 (`resolute`) codename, like the 3.0.x cells pin 1.29.0.
3. **Stale negative-test string.** `negative_provider_required.sh` matched the
   3.0 wording "unregistered scheme"; 3.5 emits "No store loader found".
   **Fix:** accept both.

HSM use is preserved — the TLS handshake signs with the azihsm key, and the
negative test confirms the key fails to load without the provider. (Provider-side
alternative evaluated and rejected: the hijack is nginx-specific and the
`fips=yes` property isn't the cause; the only provider fix would drop a tested
HSM feature.)


crates/crypto/src/libctx.rs
```diff
@@ -0,0 +1,75 @@
...
+
+static CRYPTO_LIBCTX: OnceLock<CryptoLibCtx> = OnceLock::new();
+
+fn init() -> CryptoLibCtx {
+    // A fresh OSSL_LIB_CTX has no providers; load only `default` into it.
+    // azihsm is loaded in the *process default* libctx, never in this one, so
+    // bare fetches here can never resolve to azihsm.
+    let ctx = LibCtx::new().expect("azihsm_crypto: failed to create private OSSL_LIB_CTX");
+    let default = Provider::load(Some(&ctx), "default")
+        .expect("azihsm_crypto: failed to load 'default' provider into private OSSL_LIB_CTX");
```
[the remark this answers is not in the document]
jenstopp: If creating the OSSL_LIB_CTX or loading the default provider fails, libcrypto itself is unusable and every crypto op fails regardless — there's no recoverable path. This is consistent with how the crate already handles impossible OpenSSL states (md() .unwrap(),

crates/crypto/src/hash/hash_ossl.rs
```diff
@@ -152,6 +154,27 @@ impl OsslHashAlgo {
...
+    pub(crate) fn md_name(&self) -> &'static str {
+        use openssl::nid::Nid;
+        #[allow(clippy::panic)]
+        match self.md.type_() {
+            Nid::SHA1 => "SHA1",
+            Nid::SHA256 => "SHA256",
+            Nid::SHA384 => "SHA384",
+            Nid::SHA512 => "SHA512",
+            _ => panic!("unsupported hash algorithm for libctx fetch"),
+        }
```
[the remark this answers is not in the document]
jenstopp: der_algo() directly below already panics for the same non-SHA-2 set, so the crate is already effectively limited to SHA1/256/384/512 on every signing/DER path; the only new panic case is a raw hash of an exotic digest, which isn't exercised.

crates/crypto/src/hmac/hmac_ossl.rs
```diff
@@ -10,317 +10,287 @@
...
+        } else {
+            Err(CryptoError::HmacSignUpdateError)
+        }
+    }
+
+    /// Finalises the MAC into `out` (must be at least the MAC size). Returns the
+    /// number of bytes written.
+    #[allow(unsafe_code)]
+    fn finish(&mut self, out: &mut [u8]) -> Result<usize, CryptoError> {
+        let mut outl: usize = 0;
```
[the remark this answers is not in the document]
jenstopp: fixed

crates/crypto/src/rsa/rsa_hash_sign_ossl.rs
```diff
@@ -405,21 +425,126 @@ impl OsslRsaHashSignAlgo {
...
+                // call.
+                let md = openssl::md::Md::fetch(
+                    Some(crate::libctx::crypto_libctx()),
+                    self.hash.md_name(),
+                    None,
+                )
                 .map_err(|_| CryptoError::RsaSetPropertyError)?;
+                if ffi::EVP_PKEY_CTX_set_rsa_mgf1_md(pctx, md.as_ptr()) != 1 {
+                    return Err(CryptoError::RsaSetPropertyError);
+                }
```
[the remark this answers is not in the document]
jenstopp: Not a UAF in this code path. pctx here is a provider EVP_PKEY_CTX (built via EVP_PKEY_CTX_new_from_pkey on a provider key), and for provider ctxs EVP_PKEY_CTX_set_rsa_mgf1_md forwards the digest by name — it calls EVP_MD_get0_name(md) and the provider copies that name into its own ctx and re-fetches its own EVP_MD synchronously inside this call (rsa_set_ctx_params → rsa_setup_mgf1_md). So md doesn't need to outlive the call; there's no dangling pointer at sign time. The concern would apply to a legacy (non-provider) ctx, which this never is. Strengthened the comment to make that explicit.

crates/crypto/src/hmac/hmac_ossl.rs
```diff
@@ -10,317 +10,307 @@
...
+            let bld = ffi::OSSL_PARAM_BLD_new();
+            if bld.is_null() {
+                return Err(CryptoError::HmacInitError);
+            }
+            let pushed = ffi::OSSL_PARAM_BLD_push_utf8_string(
+                bld,
+                DIGEST_PARAM.as_ptr() as *const c_char,
+                digest_name.as_ptr() as *const c_char,
+                digest_name.len() as _,
+            );
```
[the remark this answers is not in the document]
jenstopp: bsize is the string length excluding the NUL, so digest_name.len() is correct

xtask/src/integration_tests.rs
```diff
@@ -5,32 +5,147 @@
...
+                // one (which may lack newer symbols).  Prepending keeps any
+                // existing LD_LIBRARY_PATH content reachable as a fallback.
+                let existing = std::env::var("LD_LIBRARY_PATH").unwrap_or_default();
+                let new_ld = if existing.is_empty() {
+                    combined
+                } else {
+                    format!("{combined}:{existing}")
+                };
+                std::env::set_var("LD_LIBRARY_PATH", new_ld);
+            }
```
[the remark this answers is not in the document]
jenstopp: Part of the #415 (ci-matrix) commits. Would not fix until 415 is merged and this one is rebased.

crates/crypto/Cargo.toml
```diff
@@ -16,8 +16,15 @@ pem-rfc7468 = { features = ["alloc"], workspace = true }
 thiserror.workspace = true
 
 [target.'cfg(target_os = "linux")'.dependencies]
+# foreign-types provides ForeignTypeRef::as_ptr to obtain the raw OSSL_LIB_CTX*
+# for the EVP_MAC FFI in the HMAC backend. Pinned to the version openssl uses so
+# the trait impl on openssl's LibCtxRef matches.
+foreign-types = "0.3"
```
rajesh-gali: should be in workspace manifest

crates/crypto/src/ecc/ecc_ossl.rs
```diff
@@ -81,33 +79,88 @@ impl SignOp for OsslEccAlgo {
...
         &mut self,
         key: &Self::Key,
         data: &[u8],
         signature: Option<&mut [u8]>,
     ) -> Result<usize, CryptoError> {
+        use std::os::raw::c_char;
+        use std::ptr;
+
+        use foreign_types::ForeignTypeRef;
+        use openssl_sys as ffi;
```
rajesh-gali: [NIT]  avoid use statement within function scope

crates/crypto/src/ecc/ecc_ossl.rs
```diff
@@ -81,33 +79,88 @@ impl SignOp for OsslEccAlgo {
...
+                    let mut vec = vec![0u8; siglen];
+                    if ffi::EVP_PKEY_sign(
+                        ctx,
+                        vec.as_mut_ptr(),
+                        &mut siglen,
+                        data.as_ptr(),
+                        data.len() as _,
+                    ) != 1
+                    {
+                        return Err(CryptoError::EccSignError);
```
rajesh-gali: this could create a memory leak, if returned earlier without calling CTX_free, I believe we should use Drop guard for EVP_PKEY_CTX. Applicable for ECDH, RSA FFI changes.

crates/crypto/src/ecc/ecc_ossl.rs
```diff
@@ -81,33 +79,88 @@ impl SignOp for OsslEccAlgo {
...
+                let ctx = EVP_PKEY_CTX_new_from_pkey(
+                    crate::libctx::crypto_libctx_ptr(),
+                    key.pkey().as_ptr(),
+                    ptr::null(),
+                );
+                if ctx.is_null() {
+                    return Err(CryptoError::EccError);
+                }
+                let result = (|| {
+                    if ffi::EVP_PKEY_sign_init(ctx) != 1 {
```
rajesh-gali: avoid checking against magic numbers

crates/crypto/src/ecc/ecc_ossl.rs
```diff
@@ -148,12 +201,37 @@ impl VerifyOp for OsslEccAlgo {
...
+        // `EVP_PKEY_CTX_new_from_pkey` exists in OpenSSL 3.0+ libcrypto but is
+        // not bound by openssl-sys 0.9.x, so it is declared here. See
+        // [`crate::libctx`].
+        unsafe extern "C" {
+            fn EVP_PKEY_CTX_new_from_pkey(
+                libctx: *mut ffi::OSSL_LIB_CTX,
+                pkey: *mut ffi::EVP_PKEY,
+                propquery: *const c_char,
+            ) -> *mut ffi::EVP_PKEY_CTX;
+        }
```
rajesh-gali: duplicate declaration of EVP_PKEY_CTX_new_from_pkey in multiple files, I think we could move the declaration to libctx and use it

crates/crypto/src/hash/hash_ossl.rs
```diff
@@ -152,6 +154,27 @@ impl OsslHashAlgo {
...
+    /// Canonical OpenSSL name used to fetch this digest from a libctx.
+    pub(crate) fn md_name(&self) -> &'static str {
+        use openssl::nid::Nid;
+        #[allow(clippy::panic)]
+        match self.md.type_() {
+            Nid::SHA1 => "SHA1",
+            Nid::SHA256 => "SHA256",
+            Nid::SHA384 => "SHA384",
+            Nid::SHA512 => "SHA512",
+            _ => panic!("unsupported hash algorithm for libctx fetch"),
```
rajesh-gali: avoid use of panic in production code. should return error

crates/crypto/src/hmac/hmac_ossl.rs
```diff
@@ -10,317 +10,307 @@
...
-    /// The number of bytes written to the signature buffer, or the required buffer size.
-    ///
-    /// # Errors
-    ///
-    /// Returns an error if:
-    /// - `CryptoError::HmacBufferTooSmall` - Output buffer is too small
-    /// - `CryptoError::HmacSignError` - OpenSSL signing operation fails
+    /// Computes an HMAC over `data`. With `signature == None`, only returns the
+    /// required buffer size.
     fn sign(
```
rajesh-gali, line 231: function doc comments were deleted.

crates/crypto/src/hmac/hmac_ossl.rs
```diff
@@ -329,122 +319,63 @@ impl<'a> VerifyStreamingOp<'a> for OsslHmacAlgo {
...
     fn algo_mut(&mut self) -> &mut Self::Algo {
         &mut self.algo
     }
 
     /// Consumes the context and returns the underlying hash algorithm.
-    ///
-    /// # Returns
-    ///
-    /// The `OsslHash` algorithm instance.
     fn into_algo(self) -> Self::Algo {
```
rajesh-gali, line 510: doc comments are deleted through out the file

crates/crypto/src/hmac/hmac_ossl.rs
```diff
@@ -10,317 +10,307 @@
...
+
+            let bld = ffi::OSSL_PARAM_BLD_new();
+            if bld.is_null() {
+                return Err(CryptoError::HmacInitError);
+            }
+            let pushed = ffi::OSSL_PARAM_BLD_push_utf8_string(
+                bld,
+                DIGEST_PARAM.as_ptr() as *const c_char,
+                digest_name.as_ptr() as *const c_char,
+                digest_name.len() as _,
```
rajesh-gali: cast with explicit types? Similar pattern is observed in multiple files.

crates/crypto/src/kdf/hkdf_ossl.rs
```diff
@@ -153,80 +139,43 @@ impl<'a> DeriveOp for OsslHkdfAlgo<'a> {
...
+        };
+        // Extract-only yields the PRK (exactly one digest block); the expand
+        // modes yield the requested length. Reject a mismatched requested
+        // length for Extract so behavior matches the Windows (CNG) backend
+        // instead of silently ignoring `derive_len`.
+        let out_len = match self.mode {
+            HkdfMode::Extract => {
+                if derive_len != self.md.size() {
+                    return Err(CryptoError::HmacInvalidDerivedKeyLength);
+                }
```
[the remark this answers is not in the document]
jenstopp: intentional and matches the Windows (CNG) backend

rajesh-gali (dismissed):

rajesh-gali (approved):

PR #453, opened 2026-06-11
Add KBKDF

jenstopp:
> [!NOTE]
> Stacked on [#442 (openssl-3-5-fix)](https://github.com/Azure/azihsm-sdk/pull/442) — review/merge that first. This branch is based on openssl-3-5-fix, not main; the diff will only be clean once #442 is merged into main.

Adds Key-Based Key Derivation Function (NIST SP 800-108 Counter Mode, HMAC PRF) to the OpenSSL provider, mirroring the existing HKDF pattern and delegating crypto to the HSM SDK. Enables the previously commented-out KBKDF entry in the provider's KDF dispatch table.

The DDI + firmware layers already implemented KBKDF (DdiKbkdfCounterHmacDerive); this wires up the missing upper layers:

- **api/lib**: HsmKbkdfAlgo + ddi::kbkdf_derive (reuses existing TryFrom<&HsmKeyProps> for DdiKeyType)
- **api/native**: AzihsmAlgoKbkdfCounterParams + kbkdf_counter_derive_key + dispatch arm + cbindgen export
- **C provider**: azihsm_ossl_kbkdf_functions[] + enabled dispatch entry. Maps OpenSSL salt→Label / info→Context (per EVP_KDF-KB); validates mode=counter, mac=HMAC; requires at least one of Label/Context.
Tests (modeled on HKDF, added to all three existing suites): provider CLI (test_kbkdf_key_derivation), provider CAPI (kbkdf_tests.cpp), native-API CAPI (kbkdf_tests.cpp).

Compatibility: KBKDF works on OpenSSL 3.0 and 3.5 — it doesn't require 3.5, so the tests are left ungated. Validated green via act on both u2404 (3.0.13) and u2604 (3.5.5): cli + capi + nginx all pass.


plugins/ossl_prov/src/azihsm_ossl_kdf.c
```diff
@@ -907,3 +907,931 @@ const OSSL_DISPATCH azihsm_ossl_hkdf_functions[] = {
...
+    /* Deep copy IKM data */
+    dup->ikm_data = NULL;
+    if (ctx->ikm_data != NULL && ctx->ikm_data_len > 0)
+    {
+        dup->ikm_data = OPENSSL_malloc(ctx->ikm_data_len);
+        if (dup->ikm_data == NULL)
+        {
+            ERR_raise(ERR_LIB_PROV, ERR_R_MALLOC_FAILURE);
+            failed = true;
+            goto cleanup;
```
rajesh-gali, line 1174: I think early return actually frees kctx memory, which this function doesn't own. Please check.

plugins/ossl_prov/src/azihsm_ossl_kdf.c
```diff
@@ -907,3 +907,931 @@ const OSSL_DISPATCH azihsm_ossl_hkdf_functions[] = {
...
+        }
+
+        ctx->md = EVP_get_digestbyname(mdname);
+        if (ctx->md == NULL)
+        {
+            ERR_raise(ERR_LIB_PROV, PROV_R_INVALID_DIGEST);
+            return OSSL_FAILURE;
+        }
+
+        ctx->hmac_algo_id = evp_md_to_hmac_algo_id(ctx->md);
```
rajesh-gali: evp_md_to_hmac_algo_id rejects SHA1, which is accepted in API layer kbkdf_counter_derive_key

rajesh-gali (changes requested):

plugins/ossl_prov/src/azihsm_ossl_kdf.c
jenstopp: :+1:  It seems the same bug is present in HKDF, shall it be fixed as part of the PR? @rajesh-gali

jenstopp: fixed it in here as well

rajesh-gali (approved):

PR #465, opened 2026-06-21
 fw: add Uno HSM platform (PAL, drivers, SystemRDL registers)

vsonims:
Introduces the complete Uno HSM firmware platform under fw/plat/uno and
 converges the shared PAL traits, alongside a few core/crypto fixes.

 Platform (fw/plat/uno)
 - App: no_std Embassy entry point with dual-core boot trampoline.
 - PAL: UnoHsmPal implementing the converged HsmPal traits — message-driven
   IPC boot handshake, NVIC-polled driver wake loop, IO ingest/complete
   pipeline, and a DTCM core-liveliness heartbeat.
 - Drivers: aes, boot_status, core_status, gdma, iic, ipc, nvic, oic, pac,
   profile, rng, semihosting, sha, systick, uart, upka, plus the bulk_copy,
   error, single_cell, static_init, static_ref, and trace support crates.
 - Registers: SystemRDL sources (rdl/) and generated register code (reg/),
   with reggen/systemrdl tooling and an xtask workspace (build, clippy, fmt,
   copyright, audit, reggen, precheck, size/readelf helpers).
 - Memory: GSRAM and DTCM mapped to the silicon memory layout (memory.x).

 PAL traits
 - Converge the HsmPal / HsmError / HsmResult trait surface so platforms
   share one definition; update docs.

 Semihosting support (new `semihosting` app feature, off by default)
 - Gate the PAL's boot-complete SYS_READY emission behind the feature.
 - Panic handler emits an error-level trace of the PanicInfo and, under
   `semihosting`, terminates via SYS_EXIT(-1); drop the direct UART panic
   write (semihosting driver dependency is optional).
 - DualCpM7 CP_ID register (rdl + generated) backs the boot trampoline's
   core-identity check.
 - xtask build gains a `--no-default-features` passthrough.

 Core / crypto fixes
 - Fix HPKE GCM+AAD buffer overflow, x509 P-521 sizing, and a TBOR dispatch
   optimization.

 CI
 - Update Rust and CodeQL workflows.


fw/plat/uno/fw/drivers/upka/src/engine.rs
```diff
@@ -0,0 +1,430 @@
...
+    /// - `Ok(())`: Public key was written to `pub_key`.
+    /// - `Err(UpkaError::CMD_ERROR)`: Input or output buffer shape is invalid,
+    ///   or hardware rejected the command.
+    pub async fn ecc_gen_pub_key(
+        &mut self,
+        curve: UpkaEccCurve,
+        priv_key: &DmaBuf,
+        pub_key: &mut DmaBuf,
+    ) -> HsmResult<()> {
+        self.execute_cmd(
```
radutta99, line 343: Do we want to check for input size here like other functions?
`ensure_cmd_input(pub_key.len() >= point_size(curve)*2 && priv_key.len() >= hsm_point_size(curve))`

fw/plat/uno/fw/pal/src/alloc.rs
```diff
@@ -0,0 +1,483 @@
...
+    fn dma_alloc_zeroed(&self, size: usize) -> HsmResult<&mut DmaBuf> {
+        let s = self.dma_alloc(size)?;
+        s.fill(0);
+        Ok(s)
+    }
+}
+
+impl Drop for UnoScopedAlloc<'_> {
+    fn drop(&mut self) {
+        wm(self.pal, self.io_index, DMA).with(|v| *v = self.marks[1]);
```
radutta99, line 276: Question: Bumb allocated data may be secret bearing.  Do we need to make sure that the secretes are zeroized here? it will effect perf.

fw/plat/uno/fw/drivers/upka/src/engine.rs
vsonims: we should. will fix in future merges

fw/plat/uno/fw/pal/src/alloc.rs
vsonims: This is todo. We will add it in io.rs to zeroize.

radutta99 (approved):
approved.

PR #471, opened 2026-06-23
Add AES support to HSM Provider for OpenSSL 3.5

jenstopp:
Implement AES symmetric cipher support in the OpenSSL provider using the OpenSSL 3.5 EVP_SKEY opaque-symmetric-key API, so keys stay HSM-resident and no raw key material ever crosses the provider boundary.

- SKEYMGMT (azihsm_ossl_skeymgmt_aes.c): a new OSSL_OP_SKEYMGMT "AES" algorithm whose keydata wraps an opaque HSM key handle. import unmasks a masked-key blob (azihsm.masked_key); generate creates a fresh AES/AES-GCM/AES-XTS key and can persist the masked blob for later reload; export refuses SECRET_KEY so EVP_SKEY_get0_raw_key always fails.
- Cipher (azihsm_ossl_cipher.c): the stubbed AES dispatch is implemented and bound to keys through the encrypt/decrypt_skey_init hooks. AES-CBC streams through the HSM (provider-side PKCS#7 so the output fits EVP's one-block final); AES-GCM adds AEAD tag/AAD handling; AES-XTS handles the double-length key. Registers an AES-256-GCM cipher and the AES SKEYMGMT in the base provider.
- All EVP_SKEY / skey-init code is #if-guarded for OpenSSL 3.5+, so the provider still builds on 3.0 (the feature is simply absent there).
- Tests: an end-to-end AES-CBC CLI acceptance test (skeyutl -genkey + enc) and C-API gtests for CBC/GCM/XTS round-trips plus the opacity guarantee, gated to OpenSSL 3.5.


plugins/ossl_prov/integration-tests/openssl-capi/cpp/algo/cipher/aes_tests.cpp
```diff
@@ -0,0 +1,379 @@
...
+#include "utils/ossl_helpers.hpp"
+#include "utils/provider_ctx.hpp"
+
+class aes_skey : public ::testing::Test
+{
+  protected:
+    ProviderCtx prov_;
+};
+
+#if OPENSSL_VERSION_NUMBER >= 0x30500000L
```
rajesh-gali, line 34: I think we should have a run time check for openssl version, instead of compiletime conditional check.

plugins/ossl_prov/src/azihsm_ossl_skeymgmt_aes.c
```diff
@@ -0,0 +1,476 @@
...
+
+/* Create a fresh AES key inside the HSM. */
+static void *azihsm_ossl_aes_skeymgmt_generate(void *vprovctx, const OSSL_PARAM params[])
+{
+    AZIHSM_OSSL_PROV_CTX *provctx = (AZIHSM_OSSL_PROV_CTX *)vprovctx;
+    char masked_key_file[AZIHSM_MAX_FILE_PATH];
+    char kind_str[16];
+    azihsm_key_kind kind = AZIHSM_KEY_KIND_AES;
+    azihsm_algo_id keygen_algo = AZIHSM_ALGO_ID_AES_KEY_GEN;
+    size_t default_bytes = 32;
```
rajesh-gali: use macro instead of magic number to define default bytes size ?

plugins/ossl_prov/src/azihsm_ossl_skeymgmt_aes.c
```diff
@@ -0,0 +1,476 @@
...
+    {
+        const OSSL_PARAM *p;
+
+        /* Kind selector first so we know the default key length. */
+        if (azihsm_skey_get_str_param(
+                params,
+                AZIHSM_OSSL_PKEY_PARAM_KEY_KIND,
+                kind_str,
+                sizeof(kind_str)
+            ) == 1)
```
rajesh-gali, line 316: same

plugins/ossl_prov/integration-tests/openssl-capi/cpp/algo/cipher/aes_tests.cpp
jenstopp: It wouldn't compile. The test body references EVP_SKEY, EVP_CipherInit_SKEY, EVP_SKEY_generate, OSSL_SKEY_PARAM_* — symbols that don't exist in pre-3.5 headers. The provider matrix builds these tests in the OpenSSL 3.0 cells

However, there are also runtime checks: The tests are suffixed _RequiresOpenssl35 and the harness reports them ignored on 3.0 via AZIHSM_TEST_OPENSSL_MAJOR_MINOR

plugins/ossl_prov/src/azihsm_ossl_cipher.c
```diff
@@ -53,141 +350,838 @@ static int azihsm_ossl_cipher_decrypt_init(
...
+    return azihsm_ossl_cipher_get_params(params, EVP_CIPH_GCM_MODE, 32, 1, 12, 1, 1);
 }
 
-/* AES-128-XTS */
+/* AES-128-XTS: 2x key (32 bytes), 16-byte tweak, custom IV. */
 static int azihsm_ossl_aes128xts_get_params(OSSL_PARAM params[])
 {
-    return azihsm_ossl_cipher_get_params(params, EVP_CIPH_XTS_MODE, 32, 16, 16, 0);
+    return azihsm_ossl_cipher_get_params(params, EVP_CIPH_XTS_MODE, 32, 1, 16, 0, 1);
 }
```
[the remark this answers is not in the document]
jenstopp: invalid

plugins/ossl_prov/src/azihsm_ossl_cipher.c
```diff
@@ -53,141 +350,838 @@ static int azihsm_ossl_cipher_decrypt_init(
...
 
-/* AES-256-XTS */
+/* AES-256-XTS: 2x key (64 bytes), 16-byte tweak, custom IV. */
 static int azihsm_ossl_aes256xts_get_params(OSSL_PARAM params[])
 {
-    return azihsm_ossl_cipher_get_params(params, EVP_CIPH_XTS_MODE, 64, 16, 16, 0);
+    return azihsm_ossl_cipher_get_params(params, EVP_CIPH_XTS_MODE, 64, 1, 16, 0, 1);
 }
```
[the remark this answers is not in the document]
jenstopp: invalid

rajesh-gali (approved):

PR #473, opened 2026-06-23
feat(engine): typed ex_data + ENGINE_set_destroy wiring

walterchris:
Add EngineExData<T> (typed CRYPTO_get_ex_new_index slots) and Engine::set_destroy/c_destroy with the DestroyHandler trait. Adopt the &mut Engine borrow model so an outstanding &T from EngineExData::get statically conflicts with set/take detaching it (closes a safe-code use-after-free); rippled through Engine::bind and bind_helper.

### DIFF
--- a/plugins/ossl_engine/openssl-engine/src/engine.rs
+++ b/plugins/ossl_engine/openssl-engine/src/engine.rs
@@ -5,6 +5,7 @@
 
 use std::ffi::CStr;
 use std::ffi::c_char;
+use std::ffi::c_int;
 use std::ptr::NonNull;
 use std::ptr::null_mut;
 
@@ -12,7 +13,10 @@ use openssl_sys_engine as ffi;
 
 use crate::error::EngineError;
 use crate::error::EngineResult;
+use crate::error::RetCode;
+use crate::error::catch_panic;
 use crate::error::ossl_check;
+use crate::error::result_to_int;
 
 pub struct Engine {
     ptr: *mut ffi::ENGINE,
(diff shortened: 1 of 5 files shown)


plugins/ossl_engine/openssl-engine/src/engine.rs
```diff
@@ -41,10 +50,10 @@ impl Engine {
     #[allow(unsafe_code)]
     #[allow(clippy::not_unsafe_ptr_arg_deref)]
     pub unsafe fn bind(
-        &self,
+        &mut self,
         id: *const c_char,
         fns: NonNull<ffi::dynamic_fns>,
-        f: fn(&Engine, &CStr) -> EngineResult<()>,
+        f: fn(&mut Engine, &CStr) -> EngineResult<()>,
     ) -> EngineResult<()> {
```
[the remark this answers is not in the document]
walterchris: This is actually false positive. Rust allows calling &mut self methods on a temporary.

plugins/ossl_engine/openssl-engine/src/exdata.rs
```diff
@@ -0,0 +1,268 @@
...
+        assert_eq!(
+            counter.load(Ordering::SeqCst),
+            1,
+            "ENGINE_free must not re-drop after take()"
+        );
+    }
+
+    /// Pins the contract that `register` does NOT install a libcrypto-side
+    /// free callback. If this test ever fails to leak, revisit the module
+    /// docs — the caller-side `take()` contract relies on libcrypto being
```
rajesh-gali, line 243: how can we make sure every teardown safely deletes boxed memory?

rajesh-gali (approved):

plugins/ossl_engine/openssl-engine/src/exdata.rs
walterchris: The chain that frees the box:
1. Engine::set_destroy::<H>() registers c_destroy::<H> via ENGINE_set_destroy_function.
2. OpenSSL invokes it exactly once when the ENGINE is torn down (after the last ENGINE_free), synchronously, while the .so is still loaded.
3. The handler (AzihsmDestroy::destroy, in the lifecycle PR) calls slot.take(engine), which detaches and drops the Box.

PR #476, opened 2026-06-24
feat(pal): per-partition persistent store and device-global seed store

vsonims:
Add GSRAM-backed persistent storage for the Uno platform and the cross-PAL traits/derivation it relies on, so partitions can be allocated, enabled, and provisioned via Admin IPC with their secret state surviving across the partition lifecycle.

Storage drivers (Uno)
- part_store: GSRAM-backed per-partition persistent store (65 slots × 3072 B) mirroring the reference HsmPartPersistentStore byte-for-byte for shared fields; Uno-specific fields are appended into the trailing reserved region so no shared offset moves. Typed accessors hand out DmaBufs and own all reg_soc/GSRAM knowledge, keeping the PAL platform-agnostic.
- bks_store: typed accessor over the device-global BKS table (12 packed 41-byte entries: current BKS1, last-ten-SVN history, single BKS2).

Device-global seed store
- Add HsmSeedStore trait for root-of-trust seeds (mfgr/owner seeds + SVN selectors, fw_seed, uds), as an HsmPal supertrait, implemented by both the std and Uno PALs. Removes the device-global PartProps (SVN / BKS2_ID / MFGR_SEED / DEV_OWNER_SEED / UDS / FW_SEED) that did not belong in the per-partition property surface.

BK_BOOT lifecycle (key-derive crate)
- New azihsm_fw_core_crypto_key_derive crate: derive_masking_key, mask_bk_boot, unmask_bk_boot. The raw BK_BOOT is never stored — only its BKx-masked form (Masked_BK_BOOT) is created once at partition allocation and recovered on demand by unmasking, matching the reference firmware.

Uno PAL partition manager
- HsmPartitionManager backed by part_store: partition lifecycle (part_alloc / part_free / part_enable / part_disable) with rollback on failed allocation and full secret zeroization on disable and free.
- Admin IPC: SetResource (0x7f) and PfnEnableDisable (0x5) message types + handlers drive the lifecycle; the IPC layer owns dispatch and the ACK owned-table count, the partition layer owns the primitives.
- Admin IO context + admin-scoped allocator for PAL-internal DMA/crypto (identity keygen, BK_BOOT masking).

Property surface
- Store PTA_PUB_KEY as raw 96-byte x‖y coordinates (strip SEC1 0x04).
- Replace the full 167-byte policy with POLICY_HASH, the 48-byte SHA-384 digest computed in PartInit.
- Remove the now-unused u64 property surface.

DDI
- Hoist the shared recover_bk_boot helper to ddi/mod.rs so tbor no longer depends on mbor; both codecs call crate::ddi::recover_bk_boot.

RDL / reggen
- gsram_map: add the BKS table and partition-store regions; dtcm_map updates; regenerate reg/soc (part_store_t, io_gsram, hsm_dtcm).


fw/plat/uno/fw/drivers/part_store/src/part_store.rs
```diff
@@ -0,0 +1,1133 @@
...
+    // ===== Uno-specific fields, appended flat into reserved3 =====
+    // Ordered so the DMA-target public-key buffers (`ec_pub_key`,
+    // `se_pub_key`, written directly by the ECC keygen engine) land on
+    // 4-byte boundaries. The reference region ends 4-aligned, and every
+    // field up to and including the buffers has a size that is a multiple
+    // of 4, so each stays aligned; the only non-multiple-of-4 fields — the
+    // seven `[u8;2]` key handles — are placed last, before `reserved3`,
+    // where misalignment is harmless. All fields are byte arrays (align 1)
+    // so no implicit padding is introduced; accessors do the LE conversions.
+    state: [u8; 4],
```
radutta99, line 201: I am not able to imagine how we can support impactless update from the older firmware which has hsm_partition_table to some of the data fields here. Note, we also need to support update from newer firmware to older firmware in the fleet.

vsonims: We can address this in future potentially with a magic cookie.

radutta99 (approved):

PR #481, opened 2026-06-25
feat(tbor): add out-of-session PartInfo command

vsonims:
Add the TBOR `PartInfo` command (opcode 0x32): an out-of-session info command that reports device-level fields plus the bound partition's lifecycle and identity in a single round-trip. It combines the device-level fields of the MBOR `GetDeviceInfo` command with the partition identity (Partition ID and identity public key).

The response carries:
  - device_kind   (DdiDeviceKind; physical device)
  - part_state    (PartState discriminant)
  - generation    (partition generation counter)
  - owner_svn     (BKS2 selector)
  - mfgr_svn      (BKS1 selector / firmware SVN)
  - pid           (16-byte Partition ID)
  - pid_pub_key   (96-byte raw ECC-P384 identity public key)

The module-wide FIPS approval status is conveyed through the standard TBOR response header flag (set via the encode builder), not a body field, matching the existing `get_device_info` handler.

No new PAL trait methods are required — the handler reuses the existing `part_state` helpers. `PartInfo` needs no session: it is classified out-of-session in the dispatcher (not in-session, no session-id cross-check), so the default-PSK gate never applies. `handle_io` already drops IOs for non-Enabled/Initializing partitions, so the identity reads always succeed; they use `?` purely as defense-in-depth.

Wire schema is shared between firmware and host via azihsm_fw_ddi_tbor_types::part_info; the host value types live in azihsm_ddi_tbor_types::part_info and the two field orders are kept in sync so the TOC layouts match.

Tests (mirroring get_api_rev, plus a state-transition test):
  - round_trip (emu/sock)
  - part_info_repeated_stable_emu — 16-iteration stateless probe
  - part_info_independent_of_session_state_emu — stable across Pending/Active/Closed
  - part_info_reflects_part_init_transition_emu — part_state advances Enabled -> Initializing after PartInit while PID, identity public key, and SVN lineage remain stable
  - unsupported_on_mock

Promote the part_init test helpers (bootstrap_rotated_co, known_good_part_policy, mach_seed, pota_thumbprint, ROTATED_CO_PSK, CO) from pub(super) to pub(crate) so the PartInfo transition test can reuse the rotate-CO-PSK + PartInit bootstrap.

### DIFF
--- a/ddi/tbor/types/src/part_info.rs
+++ b/ddi/tbor/types/src/part_info.rs
@@ -26,8 +26,8 @@ pub const TBOR_OP_PART_INFO: u8 = 0x31;
 /// Length of the opaque partition identity blob (PID).
 pub const PID_LEN: usize = 16;
 
-/// Length of the raw ECC-P384 identity public key (`x ‖ y`, big-endian
-/// coordinates with the SEC1 `0x04` prefix stripped).
+/// Length of the raw ECC-P384 identity public key (`x ‖ y`), with each
+/// 48-byte coordinate in little-endian (HSM wire format; SEC1 `0x04` prefix stripped).
 pub const PID_PUB_KEY_LEN: usize = 96;
 
 /// Host-facing TBOR `PartInfo` request. Carries no per-call data.
--- a/ddi/tbor/types/src/part_info.rs
+++ b/ddi/tbor/types/src/part_info.rs
@@ -1,0 +1,81 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Host-side wrapper for the TBOR `PartInfo` command.
+//!
+//! `PartInfo` is an **out-of-session** info command. It combines the
+//! device-level fields of the MBOR `GetDeviceInfo` command with the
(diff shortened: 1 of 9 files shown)


radutta99 (dismissed):

ddi/tbor/types/src/part_info.rs
```diff
@@ -0,0 +1,81 @@
...
+///
+/// Field order mirrors the firmware schema in
+/// `azihsm_fw_ddi_tbor_types::part_info`; the two MUST stay in sync so
+/// the TOC layouts match.
+///
+/// The module-wide FIPS approval status is carried in the standard TBOR
+/// response header flag, not as a body field, so it is not declared
+/// here.
+#[tbor(response)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
jaygmsft, line 56: Do we need Copy in the struct?

vsonims: This is host side we should be ok to have copy.

jaygmsft (approved):

PR #501, opened 2026-06-30
feat(fw): implement GetUnwrappingKey DDI handler

jaygmsft:
Within an open session, return the partition's RSA-2048 unwrapping public key (raw wire `n_le || e_le`) plus its vault key id, for the host to RSA-AES key-wrap a payload for RsaUnwrap.

- Lazy RSA-2048 keygen via a new HsmPartitionManager:: part_ensure_unwrapping_key: the std PAL generates on first use (keygen is too slow to run at partition enable); hardware PALs generate in the background from part-init and report PendingKeyGeneration until ready, prompting the host to retry.
- Derive the public key from the vault-stored private key on demand (no separate pub cached, matching the reference firmware) via a new HsmRsa::rsa_priv_pub_key that converts the PAL's vault representation (raw on hardware, DER in the std/OpenSSL PAL) into wire form; the std driver owns the big-endian->little-endian flip.
- Serialize the public key straight into the response frame's reserved slot (query/alloc/use) — no intermediate buffer or copy.
- Centralize the BE<->LE reverse_copy helper in drivers/mod.rs, shared by the ecc and rsa drivers.
- Fix ras_gen_keypair -> rsa_gen_keypair typo.

masked_key is emitted empty for now (firmware-side masking pending the unmask path). Uno PAL methods are stubbed (PendingKeyGeneration / UnsupportedCmd).

Adds get_unwrapping_key smoke tests (happy path, stable-across-calls, no-session). Full smoke 53/53; std PAL unit 157/157.


fw/pal/traits/src/part.rs
```diff
@@ -414,6 +414,36 @@ pub trait HsmPartitionManager {
...
+    /// - Hardware PALs kick generation off in the background at
+    ///   partition init; while it is still in flight they return
+    ///   [`HsmError::PendingKeyGeneration`] so the host retries until
+    ///   the key is ready.
+    ///
+    /// # Errors
+    /// - [`HsmError::PendingKeyGeneration`] — generation is still in
+    ///   progress; the host should retry.
+    /// - Propagated PKA / vault failures.
+    async fn part_ensure_unwrapping_key(&self, io: &impl HsmIo) -> HsmResult<HsmKeyId>;
```
vsonims: Please follow the property model. We do not need to add new methods to PAL

vsonims (changes requested):

vsonims (dismissed):

vsonims (approved):

PR #519, opened 2026-07-08
fw: OpenSession + CloseSession bring-up (pub‖priv session-key split + CQE session tracking)

radutta99:
Brings up the MBOR **OpenSession** and **CloseSession** DDIs on the Uno HSM, validated on a Manticore EVB. Stacked on #515 (EstablishCredential), which has since merged, so the base is now `main`.

## Session / establish-credential ECDH key handling

The `SessionEncryption` and `EstablishCred` keys are stored as `pub(96) ‖ priv(48)` (144 B), matching the reference firmware's on-storage layout.

- `derive_session_credential_keys` (OpenSession) and `derive_credential_keys` (EstablishCredential) extract the ECDH private scalar by splitting off the leading public key (`split_at(pub_key_len)`), requiring the exact `pub ‖ priv` length — a malformed blob returns `EccInvalidKeyLength`. This replaces the earlier trailing-slice approach, which silently accepted a bare-private-key blob. Note: private-key validity (e.g. a non-zero scalar) is guaranteed at key-pair generation and is deliberately not re-validated here at consumption time.
- The **std PAL** now stores the same 144-byte `pub ‖ priv` blob for these two key kinds (`create_internal_ecc384_key`), mirroring the Uno PAL's `build_enable_key_blob`. The std firmware-equivalent `key_len` already declared 144 B for these kinds, so this also aligns the actual storage with the declared size (it previously stored only the bare 48-byte scalar).

## CQE session tracking

`handle_mbor_op` built every MBOR CQE with `session_id = None` / `session_closed = false`. The userspace emu path masks this (the mock device reads the session id from the decoded *response* header), but the Linux kernel driver tracks sessions per file handle from the *CQE* session fields (`session_ctrl_flags.{opcode,in_session_cmd,safe_to_close_session}` + `session_id`). With those unset, OpenSession never registered the session, so CloseSession failed with `FileHandleNoExistingSession (0x0870001F)`.

Populate the CQE session fields, mirroring the reference firmware (mcr-hsm `hsm_fsm::send_cqe`, which sets `id_valid = session_id.is_some()`, `session_id`, and `session_closed` on Close — the `CqeDw0` session byte is already bit-identical to mcr-hsm `HsmSessionFlags` and the kernel `session_ctrl_flags`). OpenSession surfaces its freshly-allocated id via a `DispatchResult` returned from `mbor::dispatch` (response slice + optional session id); CloseSession reuses the request header `sess_id`. Session fields are left cleared on any dispatch failure.

## Validation

**Manticore EVB:**
- `test_open_session_no_cert_chain` — EstablishCredential → GetSessionEncryptionKey → OpenSession ✓
- `test_close_session_no_cert_chain` — OpenSession then CloseSession ✓
- `test_open_session_limit` — 8 sessions open, 9th → `VaultSessionLimitReached` ✓

**Host (emu backend):** `cargo test -p azihsm_ddi_mbor_types --features emu` — OpenSession / EstablishCredential / GetSessionEncryptionKey happy paths pass; no regressions vs baseline (the only failures are pre-existing emu limitations: invalid-public-key and reopen_session cases, identical with and without this change).

## Notes
- mcr-hsm also sets `app_vault_id` (the kernel `short_app_id`) for OpenSession; it is only used by the AES fast path and is left as a follow-up.

### DIFF
--- a/fw/core/lib/src/ddi/mbor/close_session.rs
+++ b/fw/core/lib/src/ddi/mbor/close_session.rs
@@ -32,10 +32,14 @@ pub(crate) async fn close_session<'p, P: HsmPal>(
     let _body: DdiCloseSessionReq = decoder.decode_data()?;
 
     let sess_id = hdr.sess_id.ok_or(HsmError::SessionExpected)?;
-    pal.session_destroy(io, HsmSessId::from(sess_id)).await?;
 
-    // Echo the closed session id back in the response header so the
-    // host can confirm which session was torn down.
+    // Build the response BEFORE tearing down the session. The host tracks
+    // the close from the CQE `session_closed` bit, which the IO layer only
+    // sets when this handler returns `Ok`. Destroying first and then failing
+    // to encode would leave the session gone while the CQE still reports it
+    // open, desyncing the host driver's per-file-handle session tracking.
+    // Echo the closed session id back in the header so the host can confirm
+    // which session was torn down.
     let resp = pal.dma_alloc_var(io, |buf| {
         super::encode_resp(
             &super::success_hdr_sess(hdr, DdiOp::CloseSession, sess_id),
@@ -43,5 +47,8 @@ pub(crate) async fn close_session<'p, P: HsmPal>(
             buf,
         )
     })?;
+
+    pal.session_destroy(io, HsmSessId::from(sess_id)).await?;
+
     Ok(resp)
 }
--- a/fw/core/lib/src/ddi/mbor/close_session.rs
+++ b/fw/core/lib/src/ddi/mbor/close_session.rs
@@ -32,14 +32,10 @@ pub(crate) async fn close_session<'p, P: HsmPal>(
     let _body: DdiCloseSessionReq = decoder.decode_data()?;
 
     let sess_id = hdr.sess_id.ok_or(HsmError::SessionExpected)?;
+    pal.session_destroy(io, HsmSessId::from(sess_id)).await?;
 
-    // Build the response BEFORE tearing down the session. The host tracks
-    // the close from the CQE `session_closed` bit, which the IO layer only
-    // sets when this handler returns `Ok`. Destroying first and then failing
-    // to encode would leave the session gone while the CQE still reports it
-    // open, desyncing the host driver's per-file-handle session tracking.
-    // Echo the closed session id back in the header so the host can confirm
-    // which session was torn down.
+    // Echo the closed session id back in the response header so the
+    // host can confirm which session was torn down.
     let resp = pal.dma_alloc_var(io, |buf| {
         super::encode_resp(
             &super::success_hdr_sess(hdr, DdiOp::CloseSession, sess_id),
@@ -47,8 +43,5 @@ pub(crate) async fn close_session<'p, P: HsmPal>(
             buf,
         )
     })?;
-
-    pal.session_destroy(io, HsmSessId::from(sess_id)).await?;
-
     Ok(resp)
 }
--- a/fw/core/lib/src/ddi/mbor/establish_credential.rs
+++ b/fw/core/lib/src/ddi/mbor/establish_credential.rs
@@ -335,16 +335,26 @@ async fn derive_credential_keys<P: HsmPal>(
 
     let secret = pal.dma_alloc(io, HsmEccCurve::P384.secret_len())?;
     {
-        // ECDH needs the private scalar, which is the trailing
-        // `priv_key_len` bytes of the vault key. This handles both
-        // on-storage layouts: the Uno PAL stores `pub(96) ‖ priv(48)`
-        // while the std PAL stores the bare `priv(48)`.
+        // The establish-credential key is stored as
+        // `pub(pub_key_len) ‖ priv(priv_key_len)`; ECDH needs the private
+        // scalar. Split off the leading public key so a blob that is not in
+        // this layout (e.g. a bare private key with no public half) is
+        // rejected up front rather than silently keying ECDH on the wrong
+        // bytes.
+        let pub_key_len = HsmEccCurve::P384.pub_key_len();
         let priv_key_len = HsmEccCurve::P384.priv_key_len();
         let est_cred_blob = pal.vault_key(io, est_cred_key_id)?;
-        if est_cred_blob.len() < priv_key_len {
+        if est_cred_blob.len() < pub_key_len + priv_key_len {
             return Err(HsmError::InternalError);
         }
-        let (_head, priv_key) = est_cred_blob.split_at(est_cred_blob.len() - priv_key_len);
+        let (_pub_key, priv_key) = est_cred_blob.split_at(pub_key_len);
+        let priv_key = &priv_key[..priv_key_len];
+        // A zero private scalar is invalid for ECDH and indicates a corrupt
+        // or uninitialised vault slot; reject it rather than deriving a
+        // degenerate shared secret.
+        if priv_key.iter().all(|&b| b == 0) {
+            return Err(HsmError::InternalError);
+        }
         pal.ecdh_derive(
             io,
             HsmEccCurve::P384,
--- a/fw/core/lib/src/ddi/mbor/establish_credential.rs
+++ b/fw/core/lib/src/ddi/mbor/establish_credential.rs
@@ -335,20 +335,19 @@ async fn derive_credential_keys<P: HsmPal>(
 
     let secret = pal.dma_alloc(io, HsmEccCurve::P384.secret_len())?;
     {
-        // The establish-credential key is stored as
+        // The establish-credential key is stored as exactly
         // `pub(pub_key_len) ‖ priv(priv_key_len)`; ECDH needs the private
-        // scalar. Split off the leading public key so a blob that is not in
-        // this layout (e.g. a bare private key with no public half) is
-        // rejected up front rather than silently keying ECDH on the wrong
-        // bytes.
+        // scalar. Require that exact length and split off the leading public
+        // key, so a blob that is not `pub ‖ priv` — a bare private key, or one
+        // with extra trailing bytes — is rejected rather than silently keying
+        // ECDH on the wrong bytes or masking vault corruption.
         let pub_key_len = HsmEccCurve::P384.pub_key_len();
         let priv_key_len = HsmEccCurve::P384.priv_key_len();
         let est_cred_blob = pal.vault_key(io, est_cred_key_id)?;
-        if est_cred_blob.len() < pub_key_len + priv_key_len {
+        if est_cred_blob.len() != pub_key_len + priv_key_len {
             return Err(HsmError::InternalError);
         }
         let (_pub_key, priv_key) = est_cred_blob.split_at(pub_key_len);
-        let priv_key = &priv_key[..priv_key_len];
         // A zero private scalar is invalid for ECDH and indicates a corrupt
         // or uninitialised vault slot; reject it rather than deriving a
         // degenerate shared secret.
--- a/fw/core/lib/src/ddi/mbor/establish_credential.rs
+++ b/fw/core/lib/src/ddi/mbor/establish_credential.rs
@@ -345,15 +345,9 @@ async fn derive_credential_keys<P: HsmPal>(
         let priv_key_len = HsmEccCurve::P384.priv_key_len();
         let est_cred_blob = pal.vault_key(io, est_cred_key_id)?;
         if est_cred_blob.len() != pub_key_len + priv_key_len {
-            return Err(HsmError::InternalError);
+            return Err(HsmError::EccInvalidKeyLength);
         }
         let (_pub_key, priv_key) = est_cred_blob.split_at(pub_key_len);
-        // A zero private scalar is invalid for ECDH and indicates a corrupt
-        // or uninitialised vault slot; reject it rather than deriving a
-        // degenerate shared secret.
-        if priv_key.iter().all(|&b| b == 0) {
-            return Err(HsmError::InternalError);
-        }
         pal.ecdh_derive(
             io,
             HsmEccCurve::P384,
--- a/fw/core/lib/src/ddi/mbor/mod.rs
+++ b/fw/core/lib/src/ddi/mbor/mod.rs
@@ -117,6 +117,7 @@ pub(crate) async fn dispatch<'p, P: HsmPal>(
     io: &impl HsmIo,
     decoder: &mut DdiDecoder<'_>,
     hdr: &DdiReqHdr,
+    out_sess_id: &mut Option<u16>,
 ) -> HsmResult<&'p DmaBuf> {
     check_api_rev(hdr)?;
 
@@ -136,7 +137,7 @@ pub(crate) async fn dispatch<'p, P: HsmPal>(
         DdiOp::SetSealedBk3 => set_sealed_bk3(pal, io, decoder, hdr),
         DdiOp::InitBk3 => init_bk3(pal, io, decoder, hdr).await,
         DdiOp::EstablishCredential => establish_credential(pal, io, decoder, hdr).await,
-        DdiOp::OpenSession => open_session(pal, io, decoder, hdr).await,
+        DdiOp::OpenSession => open_session(pal, io, decoder, hdr, out_sess_id).await,
         DdiOp::CloseSession => close_session(pal, io, decoder, hdr).await,
         DdiOp::DeleteKey => delete_key(pal, io, decoder, hdr).await,
         DdiOp::AesGenerateKey => aes_generate_key(pal, io, decoder, hdr).await,
--- a/fw/core/lib/src/ddi/mbor/mod.rs
+++ b/fw/core/lib/src/ddi/mbor/mod.rs
@@ -167,7 +167,6 @@ pub(crate) async fn dispatch<'p, P: HsmPal>(
         DdiOp::SetSealedBk3 => set_sealed_bk3(pal, io, decoder, hdr),
         DdiOp::InitBk3 => init_bk3(pal, io, decoder, hdr).await,
         DdiOp::EstablishCredential => establish_credential(pal, io, decoder, hdr).await,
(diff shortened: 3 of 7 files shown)


fw/plat/std/pal/src/part.rs
```diff
@@ -951,16 +951,30 @@ impl StdHsmPal {
...
         let priv_len = pk.hsm_bytes_len();
         let mut priv_buf = vec![0u8; priv_len];
         pk.to_hsm_bytes(&mut priv_buf[..priv_len])
             .map_err(|_| HsmError::EccExportError)?;
 
+        // The establish-credential and session-encryption keys are stored
+        // as `pub(P384_PUB_KEY_LEN) ‖ priv(priv_len)` (144 B), matching the
+        // Uno PAL's `build_enable_key_blob` and the reference on-storage
+        // layout, so the ECDH derivation can split off the public half.
+        // Every other kind stores the bare private scalar.
```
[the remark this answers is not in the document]
radutta99: Fixed in fc0cf540. The doc comment now describes the conditional on-storage layout: `EstablishCred` and `SessionEncryption` store `pub(96) ‖ priv(48)` (144 B, matching the Uno PAL / reference layout so ECDH can split off the public half), while every other kind stores the bare raw HSM-format private scalar `d` (48 B).

fw/core/lib/src/io.rs
```diff
@@ -168,24 +168,46 @@ impl<P: HsmPal> Hsm<P> {
...
+            // per file handle: a successful OpenSession returns the newly
+            // allocated id in `session_id` (which sets `session_id_valid`), and
+            // a successful CloseSession additionally sets `session_closed`.
+            // Both are left cleared on any dispatch failure so the fields are
+            // only populated on success.
+            let (cqe_sess_id, cqe_closed) = if dispatch_result.is_ok() {
+                match session_ctrl {
+                    SessionCtrl::Open => (opened_sess_id, false),
+                    SessionCtrl::Close => (hdr.sess_id, true),
+                    _ => (None, false),
```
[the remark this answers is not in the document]
radutta99: Fixed in b3ca4f1f. CloseSession now builds the response *before* `session_destroy`. If response encoding fails, the session is left intact and the CQE reports it still open (consistent); the destroy only runs once the response is successfully built, matching the committed transition the CQE `session_closed` bit signals.

fw/core/lib/src/ddi/mbor/establish_credential.rs
```diff
@@ -335,16 +335,26 @@ async fn derive_credential_keys<P: HsmPal>(
...
+        // scalar. Split off the leading public key so a blob that is not in
+        // this layout (e.g. a bare private key with no public half) is
+        // rejected up front rather than silently keying ECDH on the wrong
+        // bytes.
+        let pub_key_len = HsmEccCurve::P384.pub_key_len();
         let priv_key_len = HsmEccCurve::P384.priv_key_len();
         let est_cred_blob = pal.vault_key(io, est_cred_key_id)?;
-        if est_cred_blob.len() < priv_key_len {
+        if est_cred_blob.len() < pub_key_len + priv_key_len {
+            return Err(HsmError::InternalError);
```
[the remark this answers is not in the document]
radutta99: Fixed in b3ca4f1f. Same exact-length enforcement here (`!= pub_key_len + priv_key_len`); longer-than-`pub‖priv` blobs are now rejected.

fw/core/lib/src/ddi/mbor/open_session.rs
```diff
@@ -132,6 +133,11 @@ pub(crate) async fn open_session<'p, P: HsmPal>(
         .session_create(io, &api_rev_bytes, mk_session, None)
         .await?;
 
+    // Surface the freshly-allocated session id so the IO layer can place it
+    // in the CQE, letting the host driver register the session against the
+    // calling file handle (required for the later CloseSession lookup).
+    *out_sess_id = Some(u16::from(sess_id));
+
```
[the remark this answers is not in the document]
radutta99: Fixed in b3ca4f1f. OpenSession now wraps `encode_response`: on any post-create failure it best-effort `session_destroy(sess_id)` (and clears the CQE out-param) before returning the error, so the committed session slot can't leak.

fw/core/lib/src/ddi/mbor/open_session.rs
```diff
@@ -210,7 +216,26 @@ async fn derive_session_credential_keys<P: HsmPal>(
...
+        // rejected up front rather than silently keying ECDH on the wrong
+        // bytes.
+        let pub_key_len = HsmEccCurve::P384.pub_key_len();
+        let priv_key_len = HsmEccCurve::P384.priv_key_len();
+        let sess_enc_blob = pal.vault_key(io, sess_enc_key_id)?;
+        if sess_enc_blob.len() < pub_key_len + priv_key_len {
+            return Err(HsmError::InternalError);
+        }
+        let (_pub_key, priv_key) = sess_enc_blob.split_at(pub_key_len);
+        let priv_key = &priv_key[..priv_key_len];
```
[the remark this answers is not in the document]
radutta99: Fixed in b3ca4f1f. The guard is now an *exact*-length check (`!= pub_key_len + priv_key_len`). A `pub‖priv‖junk` blob is rejected rather than accepted with trailing bytes silently ignored.

fw/plat/std/pal/src/part.rs
```diff
@@ -951,16 +956,30 @@ impl StdHsmPal {
...
+        let key_blob: Vec<u8> = match kind {
+            HsmVaultKeyKind::EstablishCred | HsmVaultKeyKind::SessionEncryption => {
+                let mut blob = Vec::with_capacity(P384_PUB_KEY_LEN + priv_len);
+                blob.extend_from_slice(&pub_key_out[..P384_PUB_KEY_LEN]);
+                blob.extend_from_slice(&priv_buf[..priv_len]);
```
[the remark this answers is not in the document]
radutta99: Fixed in ce364f7e. Both transient copies are now zeroized with best-effort `fill(0)` once the vault owns its copy (mirroring `provision_unwrapping_key`): `priv_buf` and the `key_blob` copy. `core::mem::take` moves the scalar into `key_blob` for the bare-priv case so both buffers scrub uniformly, and the scrub runs on the error path too.

fw/core/lib/src/ddi/mbor/close_session.rs
```diff
@@ -32,16 +32,23 @@ pub(crate) async fn close_session<'p, P: HsmPal>(
     let _body: DdiCloseSessionReq = decoder.decode_data()?;
 
     let sess_id = hdr.sess_id.ok_or(HsmError::SessionExpected)?;
-    pal.session_destroy(io, HsmSessId::from(sess_id)).await?;
 
-    // Echo the closed session id back in the response header so the
-    // host can confirm which session was torn down.
+    // Build the response BEFORE tearing down the session. The host tracks
+    // the close from the CQE `session_closed` bit, which the IO layer only
```
jaygmsft: This is not correct and not needed. If you track the possible failures those are not relevant as those conditions are taken care of. More importantly, a failure can happen after before the response when the parent layer tries to send the response. That is the failure you wouldnt be protected against today. There is a undo log feature in the works that will take care of that at which point this will be moot too.

fw/core/lib/src/ddi/mbor/establish_credential.rs
```diff
@@ -335,16 +335,25 @@ async fn derive_credential_keys<P: HsmPal>(
...
+        // scalar. Require that exact length and split off the leading public
+        // key, so a blob that is not `pub ‖ priv` — a bare private key, or one
+        // with extra trailing bytes — is rejected rather than silently keying
+        // ECDH on the wrong bytes or masking vault corruption.
+        let pub_key_len = HsmEccCurve::P384.pub_key_len();
         let priv_key_len = HsmEccCurve::P384.priv_key_len();
         let est_cred_blob = pal.vault_key(io, est_cred_key_id)?;
-        if est_cred_blob.len() < priv_key_len {
+        if est_cred_blob.len() != pub_key_len + priv_key_len {
+            return Err(HsmError::InternalError);
```
jaygmsft: Can we do better error codes please instead of internal error?

fw/core/lib/src/ddi/mbor/establish_credential.rs
```diff
@@ -335,16 +335,25 @@ async fn derive_credential_keys<P: HsmPal>(
...
+        }
+        let (_pub_key, priv_key) = est_cred_blob.split_at(pub_key_len);
+        // A zero private scalar is invalid for ECDH and indicates a corrupt
+        // or uninitialised vault slot; reject it rather than deriving a
+        // degenerate shared secret.
+        if priv_key.iter().all(|&b| b == 0) {
```
jaygmsft: When will this case happen? I am not sure if this check is needed.

jaygmsft: Also iter().all will do byte by byte comparison. Do we need constant time here?

fw/core/lib/src/ddi/mbor/mod.rs
```diff
@@ -117,6 +117,7 @@ pub(crate) async fn dispatch<'p, P: HsmPal>(
     io: &impl HsmIo,
     decoder: &mut DdiDecoder<'_>,
     hdr: &DdiReqHdr,
+    out_sess_id: &mut Option<u16>,
```
jaygmsft: We seem to be returning one value via out parameter and one via Result, why the discrepancy?

fw/core/lib/src/ddi/mbor/open_session.rs
```diff
@@ -63,6 +63,7 @@ pub(crate) async fn open_session<'p, P: HsmPal>(
     io: &impl HsmIo,
     decoder: &mut DdiDecoder<'_>,
     hdr: &DdiReqHdr,
+    out_sess_id: &mut Option<u16>,
```
jaygmsft: We seem to be returning one value via out parameter and one via Result, why the discrepancy?

fw/core/lib/src/ddi/mbor/open_session.rs
```diff
@@ -132,8 +133,18 @@ pub(crate) async fn open_session<'p, P: HsmPal>(
         .session_create(io, &api_rev_bytes, mk_session, None)
         .await?;
 
+    // Surface the freshly-allocated session id so the IO layer can place it
+    // in the CQE, letting the host driver register the session against the
+    // calling file handle (required for the later CloseSession lookup).
+    *out_sess_id = Some(u16::from(sess_id));
```
jaygmsft: I think this can be after the encode response so that you dont have it set it case of error from it.

fw/core/lib/src/ddi/mbor/open_session.rs
```diff
@@ -143,7 +154,15 @@ pub(crate) async fn open_session<'p, P: HsmPal>(
         crate::part_state::part_mfgr_svn(pal),
         u16::try_from(crate::part_state::part_owner_svn(pal)).map_err(|_| HsmError::InvalidArg)?,
     )
-    .await?;
+    .await
+    {
+        Ok(resp) => resp,
+        Err(e) => {
+            *out_sess_id = None;
```
jaygmsft: We have undo log in the plan and its best to handle this there rather than piece meal something.

fw/plat/std/pal/src/part.rs
```diff
@@ -951,16 +956,42 @@ impl StdHsmPal {
...
+        // the scalar out for the bare case (leaving `priv_buf` empty) so both
+        // buffers can be scrubbed uniformly below.
+        let mut key_blob: Vec<u8> = match kind {
+            HsmVaultKeyKind::EstablishCred | HsmVaultKeyKind::SessionEncryption => {
+                let mut blob = Vec::with_capacity(P384_PUB_KEY_LEN + priv_len);
+                blob.extend_from_slice(&pub_key_out[..P384_PUB_KEY_LEN]);
+                blob.extend_from_slice(&priv_buf[..priv_len]);
+                blob
+            }
+            _ => core::mem::take(&mut priv_buf),
```
jaygmsft: I think this snippet can be simplified and cleaned up.

fw/plat/std/pal/src/part.rs
```diff
@@ -951,16 +956,42 @@ impl StdHsmPal {
...
+        let result = {
+            let table = self.table_mut();
+            let entry = &mut table.entries[pid as usize];
+            entry.vault.create(&key_blob, kind, None, attrs)
+        };
+
+        // Scrub the transient private-key copies from the heap now that the
+        // vault owns its own copy (best-effort `fill(0)`, mirroring
+        // `provision_unwrapping_key`). Runs on the error path too.
+        priv_buf.fill(0);
```
jaygmsft: 2 things, havent you zeroized priv_buf already in the else case in in line 979? It will be twice in that scenario. Also fill can be optimized by compiler and if you must then please use zeroize crate.

jaygmsft (changes requested):

fw/core/lib/src/ddi/mbor/open_session.rs
radutta99: I agree. We can remove this when undo log is here? Is your suggestion now to remove this code?

jaygmsft: Lets not have it for now as people forget to clean up.

fw/core/lib/src/ddi/mbor/close_session.rs
radutta99: Agreed — close_session  is back to destroy-then-encode, leaving the transactional/parent-layer cleanup to the undo log.

fw/core/lib/src/ddi/mbor/establish_credential.rs
radutta99: now returns  EccInvalidKeyLength

fw/core/lib/src/ddi/mbor/establish_credential.rs
radutta99: Removed the zero-scalar check

fw/core/lib/src/ddi/mbor/open_session.rs
radutta99: the out param is not gone.

fw/core/lib/src/ddi/mbor/open_session.rs
radutta99: the out-param is removed

fw/core/lib/src/ddi/mbor/open_session.rs
radutta99: Agreed. Removed.

fw/plat/std/pal/src/part.rs
radutta99: Agreed with both. Modified

fw/core/lib/src/ddi/mbor/open_session.rs
```diff
@@ -210,7 +217,19 @@ async fn derive_session_credential_keys<P: HsmPal>(
...
+        // key, so a blob that is not `pub ‖ priv` — a bare private key, or one
+        // with extra trailing bytes — is rejected rather than silently keying
+        // ECDH on the wrong bytes or masking vault corruption.
+        let pub_key_len = HsmEccCurve::P384.pub_key_len();
+        let priv_key_len = HsmEccCurve::P384.priv_key_len();
+        let sess_enc_blob = pal.vault_key(io, sess_enc_key_id)?;
+        if sess_enc_blob.len() != pub_key_len + priv_key_len {
+            return Err(HsmError::EccInvalidKeyLength);
+        }
+        let (_pub_key, priv_key) = sess_enc_blob.split_at(pub_key_len);
```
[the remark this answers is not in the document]
radutta99: Intentional \u2014 the PR description has been updated. We don't validate the private scalar at consumption time; a valid (non-zero) key is guaranteed at key-pair generation. This path only enforces the exact `pub\u2016priv` length (`EccInvalidKeyLength` otherwise).

fw/core/lib/src/ddi/mbor/mod.rs
```diff
@@ -136,7 +167,7 @@ pub(crate) async fn dispatch<'p, P: HsmPal>(
         DdiOp::SetSealedBk3 => set_sealed_bk3(pal, io, decoder, hdr),
         DdiOp::InitBk3 => init_bk3(pal, io, decoder, hdr).await,
         DdiOp::EstablishCredential => establish_credential(pal, io, decoder, hdr).await,
-        DdiOp::OpenSession => open_session(pal, io, decoder, hdr).await,
+        DdiOp::OpenSession => unreachable!("OpenSession is handled before this match"),
```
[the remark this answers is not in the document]
radutta99: Fixed in 676405a7. `OpenSession` is handled by an early `return` before the match, so the arm was removed \u2014 an unexpected `OpenSession` now falls through to the existing `_ => UnsupportedCmd` error instead of panicking on a trust boundary.

fw/core/lib/src/ddi/mbor/establish_credential.rs
```diff
@@ -335,16 +335,19 @@ async fn derive_credential_keys<P: HsmPal>(
...
+        let pub_key_len = HsmEccCurve::P384.pub_key_len();
         let priv_key_len = HsmEccCurve::P384.priv_key_len();
         let est_cred_blob = pal.vault_key(io, est_cred_key_id)?;
-        if est_cred_blob.len() < priv_key_len {
-            return Err(HsmError::InternalError);
+        if est_cred_blob.len() != pub_key_len + priv_key_len {
+            return Err(HsmError::EccInvalidKeyLength);
         }
-        let (_head, priv_key) = est_cred_blob.split_at(est_cred_blob.len() - priv_key_len);
+        let (_pub_key, priv_key) = est_cred_blob.split_at(pub_key_len);
```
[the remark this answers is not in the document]
radutta99: Same as OpenSession \u2014 private-key validity is ensured at key-pair generation, not re-checked at consumption. PR description updated to match; this path enforces the exact `pub\u2016priv` length only.

jaygmsft (approved):

PR #540, opened 2026-07-13
feat(mbor): implement AttestKey command handler

vsonims:
## Summary

Implements the MBOR **`AttestKey`** command (`DdiOp::AttestKey` = 1016) in the firmware. Given a vault `key_id` + caller `report_data`, it produces a tagged COSE_Sign1 key-attestation report over that key, **ES384-signed by the partition-identity (PID) key** — the same signer used for the other firmware key-attestation reports.

The wire types (`DdiAttestKeyReq`/`Resp`), opcode, and a comprehensive test file already existed; this PR adds the missing firmware handler + dispatch, plus a derived-key provenance alignment the handler exposed.

## Handler (`fw/core/lib/src/ddi/mbor/attest_key.rs`)

- Attests **ECC-private** (P-256/384/521) and **RSA-private** (2k/3k/4k, plain + CRT) keys — the public component is derived on-device (`ecc_priv_pub_key` / `rsa_priv_pub_key`) and embedded as an RFC 9052 COSE_Key.
- **Symmetric / bulk / HMAC / secret** keys are attestable too, with an empty COSE_Key (matching the reference firmware / simulator).
- **No RSA verification** — the report signature is always ES384 over the PID key; RSA only appears as embedded public material.
- Pure assembly of existing pieces: reuses the `azihsm_fw_core_crypto_key_report` encoder + existing PAL calls. **No new PAL methods, no new crates.**
- Public coordinates / modulus are reversed to big-endian before building the COSE_Key, since they are baked into the opaque signed `report` blob (which the MBOR wire layer does not convert — typed fields rely on `pre_encode`/`post_decode`, the report is the exception).

## Derived-key provenance alignment

The firmware marked on-device-**derived** keys `local=false` (strict PKCS#11 `CKA_LOCAL`), so they attested as *imported*. The simulator/reference and the existing masked-key tests treat them as **created on-device**. Since the **only** firmware reader of `local` is the attestation flag mapping, this PR aligns:

- HKDF/KBKDF-derived **HMAC** (`for_var_hmac`), **ECDH-derived secrets** (`for_ecdh_secret`), and HKDF/KBKDF-derived **AES** (`for_aes(..., true)`) → `local=true`.
- Imported / unwrapped keys (RsaUnwrap) stay non-`local`.

This makes the firmware consistent with the simulator **and** the pre-existing `secret_hkdf_derive` / `secret_kbkdf_derive` / `masked_key_*` tests, which already assert `MaskedKeyAttributes::LOCAL` for derived keys.

## Tests

- **`attest_key_smoke.rs`** — emu-runnable ECC round-trip (verifies the ES384/PID report against the partition cert chain and round-trips the attested key), in a separate `*_smoke.rs` file per convention.
- **`test_attest_hkdf_derived_aes_key`** / **`test_attest_kbkdf_derived_aes_key`** — lock in that derived AES keys attest as *generated* with an empty COSE_Key.

## Validation

- fw core + uno build; clippy `-D warnings`, nightly rustfmt (root + uno), copyright: all clean.
- **emu**: ECC×2, RSA×2, AES, HMAC, secret, HKDF-AES, KBKDF-AES attest tests pass, plus the smoke test.
- **mock** (sim, unaffected by this fw-only change): all attest tests pass.

### Known out-of-scope emu failures (not this handler)
- `test_attest_aes_xts_bulk_key` — AES-XTS not implemented yet.
- 6 `test_attest_masked_key_rsa_*` — emu masked-key placeholder (fail before attest). Both are HW/sim-covered and excluded from the smoke-filtered emu CI.

### DIFF
--- a/fw/core/lib/src/ddi/mbor/attest_key.rs
+++ b/fw/core/lib/src/ddi/mbor/attest_key.rs
@@ -1,0 +1,210 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! DDI `AttestKey` command handler.
+//!
+//! Within an open session, produce a tagged COSE_Sign1 key-attestation
+//! report over a vault key identified by `key_id`, and return it. The
+//! report is signed **ES384 by the partition-identity (PID) key** — the
+//! same signer the firmware uses for the other key-attestation reports.
+//!
+//! ECC-private and RSA-private keys embed their public component in the
+//! report as an RFC 9052 COSE_Key. Every other kind (symmetric / bulk /
+//! HMAC / secret, or any key with no public component) is still
+//! attestable, but carries an **empty** COSE_Key (`public_key_size == 0`)
+//! — matching the reference firmware and simulator.
+//!
+//! Deriving the ECC / RSA public component uses the PAL's
+//! `ecc_priv_pub_key` / `rsa_priv_pub_key`. These are implemented on the
+//! std PAL (and exercised by the emu tests); on the Uno PAL they are still
+//! stubs pending PKA bring-up, so ECC / RSA attestation returns
+//! `UnsupportedCmd` on Uno hardware until those methods land. Symmetric
+//! attestation (empty COSE_Key) needs no public-key derivation and works
+//! on any PAL.
+//!
+//! The command **persists nothing** — it reads the attested key, derives
+//! its public component, and returns a signed report. No RSA verification
+//! is performed anywhere: the report signature is always ES384 over the
+//! PID key regardless of the attested key type.
+//!
+//! ## Endianness
+//!
+//! The attested key's public coordinates / modulus are baked into the
+//! opaque signed `report` byte array, which the MBOR wire layer does not
+//! convert. The PAL emits public keys in wire-LE, so this handler reverses
+//! each component to **big-endian** before building the COSE_Key (RFC 9052
+//! uses big-endian). Typed MBOR fields would instead rely on the wire
+//! layer's `pre_encode`/`post_decode`; the report is the exception because
+//! it is a signed opaque blob.
+
+use azihsm_fw_core_crypto_key_report::key_report;
+use azihsm_fw_core_crypto_key_report::AttestedPubKey;
+use azihsm_fw_core_crypto_key_report::KeyFlags;
+use azihsm_fw_core_crypto_key_report::KeyReportParams;
+use azihsm_fw_core_crypto_key_report::APP_UUID_LEN;
+use azihsm_fw_ddi_mbor_types::attest_key::DdiAttestKeyReq;
+use azihsm_fw_ddi_mbor_types::attest_key::DdiAttestKeyResp;
+
+use super::*;
+
+/// Translate the attested key's vault attributes into report [`KeyFlags`].
+fn key_flags_from_attrs(attrs: HsmVaultKeyAttrs) -> KeyFlags {
+    KeyFlags::new()
+        .with_is_imported(!attrs.local())
+        .with_is_session_key(attrs.session())
+        .with_is_generated(attrs.local())
+        .with_can_encrypt(attrs.encrypt())
+        .with_can_decrypt(attrs.decrypt())
+        .with_can_sign(attrs.sign())
+        .with_can_verify(attrs.verify())
+        .with_can_wrap(attrs.wrap())
+        .with_can_unwrap(attrs.unwrap())
+        .with_can_derive(attrs.derive())
+}
+
+/// Reverse-copy `src` into `dst[..src.len()]` (wire-LE → big-endian).
+fn reverse_copy(dst: &mut [u8], src: &[u8]) {
+    for (d, s) in dst.iter_mut().zip(src.iter().rev()) {
+        *d = *s;
+    }
+}
+
+/// Handle `DdiAttestKeyCmd`.
+pub(crate) async fn attest_key<'p, P: HsmPal>(
+    pal: &'p P,
+    io: &impl HsmIo,
+    decoder: &mut DdiDecoder<'_>,
+    hdr: &DdiReqHdr,
+) -> HsmResult<&'p DmaBuf> {
+    let body: DdiAttestKeyReq = decoder.decode_data()?;
+    let sess_id = hdr.sess_id.ok_or(HsmError::SessionExpected)?;
+    let key_id = HsmKeyId::from(body.key_id);
+
+    // Attested key metadata → report flags.
+    let kind = pal.vault_key_kind(io, key_id)?;
+    let attrs = pal.vault_key_attrs(io, key_id)?;
+    let flags: u32 = key_flags_from_attrs(attrs).into();
+
+    // `report_data` borrows directly from the decoded request buffer — no
+    // copy; `key_report` validates its exact `REPORT_DATA_LEN` length.
+    //
+    // `app_uuid` / `vm_launch_id` are IO-lifetime allocations: they feed
+    // both `key_report` passes, so they must live outside any scoped
+    // allocator whose bump-mark reset could overlap the response buffer.
+    let app_uuid = pal.dma_alloc(io, APP_UUID_LEN)?;
+    app_uuid.copy_from_slice(&crate::session::session_app_id(
+        pal,
+        io,
+        HsmSessId::from(sess_id),
+    )?);
+
+    let vm_launch_id = crate::part_state::part_vm_launch_guid(pal, io)?;
+
+    // Derive the attested key's public component in big-endian COSE form.
+    // The private key material is fetched only in the ECC/RSA arms; the
+    // symmetric path has no public component and never touches it.
+    let key = if let Ok(curve) = super::from_pal::ecc_curve(kind) {
+        let priv_key = pal.vault_key(io, key_id)?;
+        let coord_raw = curve.priv_key_len();
+        let coord_wire = curve.wire_coord_len();
+        let wire_len = curve.wire_pub_key_len();
+        let pub_le = pal.dma_alloc(io, wire_len)?;
+        // The serialize pass must write the full wire key; a short write
+        // would leave stale DMA bytes in the tail of `pub_le` that must not
+        // be baked into the report — treat any mismatch as a PAL fault.
+        if pal
+            .ecc_priv_pub_key(io, priv_key, Some(&mut *pub_le))
+            .await?
+            != wire_len
+        {
+            return Err(HsmError::InternalError);
+        }
+        let x_be = pal.dma_alloc(io, coord_raw)?;
+        let y_be = pal.dma_alloc(io, coord_raw)?;
+        reverse_copy(x_be, &pub_le[..coord_raw]);
+        reverse_copy(y_be, &pub_le[coord_wire..coord_wire + coord_raw]);
+        AttestedPubKey::Ecc {
+            curve,
+            x: x_be,
+            y: y_be,
+        }
+    } else if let Ok(rsa) = super::from_pal::rsa_key(kind) {
+        let priv_key = pal.vault_key(io, key_id)?;
+        let mod_len = rsa.modulus_len();
+        // The PAL emits the RSA public key as wire-LE `n_le || e_le`, where
+        // `e_le` is a fixed 4-byte little-endian exponent, so the wire is
+        // exactly `mod_len + RSA_PUB_EXP_LEN`. Any other length means the
+        // PAL and our modulus-kind mapping disagree — a PAL/internal fault,
+        // not a caller error.
+        const RSA_PUB_EXP_LEN: usize = 4;
+        let wire = mod_len + RSA_PUB_EXP_LEN;
+        if pal.rsa_priv_pub_key(io, priv_key, None)? != wire {
+            return Err(HsmError::InternalError);
+        }
+        let pub_le = pal.dma_alloc(io, wire)?;
+        // The serialize pass must write exactly the queried length; a
+        // shorter write would leave stale bytes in the tail of `pub_le`.
+        if pal.rsa_priv_pub_key(io, priv_key, Some(&mut *pub_le))? != wire {
+            return Err(HsmError::InternalError);
+        }
+        let n_be = pal.dma_alloc(io, mod_len)?;
+        let e_be = pal.dma_alloc(io, RSA_PUB_EXP_LEN)?;
+        reverse_copy(n_be, &pub_le[..mod_len]);
+        reverse_copy(e_be, &pub_le[mod_len..wire]);
+        AttestedPubKey::Rsa { n: n_be, e: e_be }
+    } else {
+        // Any other kind (symmetric / bulk / HMAC / secret) is attestable
+        // but has no public component: emit an empty COSE_Key, matching the
+        // reference firmware. The report still binds the key's flags and
+        // report_data under the PID signature.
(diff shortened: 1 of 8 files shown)


fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,199 @@
...
+//!
+//! Within an open session, produce a tagged COSE_Sign1 key-attestation
+//! report over a vault key identified by `key_id`, and return it. The
+//! report is signed **ES384 by the partition-identity (PID) key** — the
+//! same signer the firmware uses for the other key-attestation reports.
+//!
+//! Both ECC-private and RSA-private keys are attestable; the attested
+//! key's public component is embedded in the report as an RFC 9052
+//! COSE_Key. Every other kind (symmetric, public-only, internal) is
+//! rejected with [`HsmError::InvalidKeyType`].
```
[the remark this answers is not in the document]
vsonims: Fixed the module docs to match the behavior: ECC/RSA-private keys embed their public component as a COSE_Key, and every other kind (symmetric / bulk / HMAC / secret) is still attestable but carries an **empty** COSE_Key (`public_key_size == 0`). No kind is rejected with `InvalidKeyType` — that stale wording is removed.

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,199 @@
...
+        })
+        .await?;
+    if report_len > DdiAttestKeyResp::MAX_REPORT_SIZE {
+        return Err(HsmError::InternalError);
+    }
+    let report = pal.dma_alloc(io, report_len)?;
+    pal.alloc_scoped_async(io, async |a| {
+        key_report(pal, io, a, &params, pid_priv, Some(&mut *report)).await
+    })
+    .await?;
```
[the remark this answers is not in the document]
vsonims: Good catch. The copy pass now captures `key_report`'s returned `written` length and rejects a short write with `InternalError` (`if written != report_len`), so stale/uninitialized DMA bytes can't be returned to the host.

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,199 @@
...
+    let (resp, layout) = pal.dma_alloc_var_with(io, |buf| {
+        let mut encoder = super::encode_resp_hdr(
+            &super::success_hdr_sess(hdr, DdiOp::AttestKey, sess_id),
+            buf,
+        )?;
+        let layout = DdiAttestKeyResp::reserve(&mut encoder, report_len)?;
+        Ok((encoder.position(), layout))
+    })?;
+    let frame = DdiAttestKeyResp::from_layout(resp, &layout);
+    frame.report.copy_from_slice(&report[..report_len]);
```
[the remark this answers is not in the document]
vsonims: Done — reworked to zero-copy: the response frame is reserved first, then `key_report`'s copy pass writes directly into the reserved `report` slot (`Some(&mut *frame.report)`). The intermediate `report` buffer and the extra `copy_from_slice` are gone.

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,207 @@
...
+        if wire <= mod_len {
+            return Err(HsmError::InvalidArg);
+        }
+        let pub_le = pal.dma_alloc(io, wire)?;
+        pal.rsa_priv_pub_key(io, priv_key, Some(&mut *pub_le))?;
+        let n_be = pal.dma_alloc(io, mod_len)?;
+        let e_be = pal.dma_alloc(io, wire - mod_len)?;
+        reverse_copy(n_be, &pub_le[..mod_len]);
+        reverse_copy(e_be, &pub_le[mod_len..wire]);
+        AttestedPubKey::Rsa { n: n_be, e: e_be }
```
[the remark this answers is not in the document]
vsonims: Confirmed — `ecc_priv_pub_key` / `rsa_priv_pub_key` are still `UnsupportedCmd` stubs on the Uno PAL (the TODOs you cited, pending PKA bring-up for RsaUnwrap/import). ECC/RSA `AttestKey` works on the std PAL (exercised by the emu tests); on Uno it returns `UnsupportedCmd` until those methods land, while **symmetric** attestation (empty COSE_Key, no pub derivation) works on any PAL. The handler uses the standard PAL abstraction correctly — this is a tracked Uno bring-up gap, not an AttestKey bug, so I've kept it scoped and documented it explicitly in the handler module docs rather than shipping untested Uno PKA code (there's no Uno test path in the MBOR emu suite to validate it). Uno implementation will follow with the rest of the Uno pub-derive bring-up.

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,207 @@
...
+            curve,
+            x: x_be,
+            y: y_be,
+        }
+    } else if let Some(mod_len) = rsa_modulus_len_for_kind(kind) {
+        // The PAL emits the RSA public key as wire-LE `n_le || e_le`.
+        let wire = pal.rsa_priv_pub_key(io, priv_key, None)?;
+        if wire <= mod_len {
+            return Err(HsmError::InvalidArg);
+        }
```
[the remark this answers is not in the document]
vsonims: Fixed — changed to `HsmError::InternalError` (a wire length ≤ modulus is an impossible PAL result, not a caller error).

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,218 @@
...
+        // A wire length no larger than the modulus is impossible (there is
+        // no room for the exponent) — treat it as a PAL/internal fault, not
+        // a caller error.
+        if wire <= mod_len {
+            return Err(HsmError::InternalError);
+        }
+        let pub_le = pal.dma_alloc(io, wire)?;
+        pal.rsa_priv_pub_key(io, priv_key, Some(&mut *pub_le))?;
+        let n_be = pal.dma_alloc(io, mod_len)?;
+        let e_be = pal.dma_alloc(io, wire - mod_len)?;
```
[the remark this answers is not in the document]
vsonims: Tightened the RSA wire-length contract as requested. The exponent is a fixed 4-byte LE field (`RSA_PUB_EXP_WIRE_LEN = 4`), so the wire is exactly `mod_len + 4`. The query pass now checks `!= mod_len + RSA_PUB_EXP_LEN` (exact) and returns `InternalError` on any mismatch, and the serialize (`Some`) pass return value is likewise checked against the same expected length, mapping a short/inconsistent write to `InternalError` before any bytes are read. `e_be` is now sized to the fixed `RSA_PUB_EXP_LEN` rather than `wire - mod_len`, so no stale-byte or `InvalidArg`-surfacing path remains. Fixed in f27b7a72.

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,224 @@
...
+    let vm_launch_id = crate::part_state::part_vm_launch_guid(pal, io)?;
+
+    // Derive the attested key's public component in big-endian COSE form.
+    let priv_key = pal.vault_key(io, key_id)?;
+    let key = if let Some(curve) = ecc_curve_for_kind(kind) {
+        let coord_raw = curve.priv_key_len();
+        let coord_wire = curve.wire_coord_len();
+        let pub_le = pal.dma_alloc(io, curve.wire_pub_key_len())?;
+        pal.ecc_priv_pub_key(io, priv_key, Some(&mut *pub_le))
+            .await?;
```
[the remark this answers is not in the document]
vsonims: Applied the same defense to the ECC path. `ecc_priv_pub_key` follows the same query/alloc/use contract and returns the wire length (`wire_pub_key_len()`), so the serialize (`Some`) pass return value is now checked against the allocated `wire_len` and any mismatch (short write) returns `InternalError` before the coordinates are read from `pub_le` — so no stale DMA tail can be baked into the report. Fixed in c9493d11.

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,233 @@
...
+use azihsm_fw_core_crypto_key_report::APP_UUID_LEN;
+use azihsm_fw_core_crypto_key_report::REPORT_DATA_LEN;
+use azihsm_fw_ddi_mbor_types::attest_key::DdiAttestKeyReq;
+use azihsm_fw_ddi_mbor_types::attest_key::DdiAttestKeyResp;
+
+use super::*;
+
+/// Map an ECC-private vault key kind to its NIST curve, or `None` if the
+/// kind is not an ECC private key.
+fn ecc_curve_for_kind(kind: HsmVaultKeyKind) -> Option<HsmEccCurve> {
```
jaygmsft: I think method exists in other handlers or places. Please reuse.

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,233 @@
...
+        HsmVaultKeyKind::Ecc256Private => Some(HsmEccCurve::P256),
+        HsmVaultKeyKind::Ecc384Private => Some(HsmEccCurve::P384),
+        HsmVaultKeyKind::Ecc521Private => Some(HsmEccCurve::P521),
+        _ => None,
+    }
+}
+
+/// Map an RSA-private vault key kind (plain or CRT) to its modulus length
+/// in bytes, or `None` if the kind is not an RSA private key.
+fn rsa_modulus_len_for_kind(kind: HsmVaultKeyKind) -> Option<usize> {
```
jaygmsft: Please check if one of the existing method can be reused or not.

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,233 @@
...
+        .with_can_decrypt(attrs.decrypt())
+        .with_can_sign(attrs.sign())
+        .with_can_verify(attrs.verify())
+        .with_can_wrap(attrs.wrap())
+        .with_can_unwrap(attrs.unwrap())
+        .with_can_derive(attrs.derive())
+}
+
+/// Reverse-copy `src` into `dst[..src.len()]` (wire-LE → big-endian).
+fn reverse_copy(dst: &mut [u8], src: &[u8]) {
```
jaygmsft, line 66: I think method exists in other handlers or places. Please reuse.

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    // Attested key metadata → report flags.
+    let kind = pal.vault_key_kind(io, key_id)?;
+    let attrs = pal.vault_key_attrs(io, key_id)?;
+    let flags: u32 = key_flags_from_attrs(attrs).into();
+
+    // Report payload fields. These must be IO-lifetime (they feed both
+    // `key_report` passes) and allocated outside any scoped allocator so a
+    // scope's bump-mark reset can't overlap the response buffer.
+    let report_data_src = &*body.report_data;
```
jaygmsft: Do we need &*? Should be able to assign directly.

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,233 @@
...
+    if report_data_src.len() != REPORT_DATA_LEN {
+        return Err(HsmError::InvalidArg);
+    }
+    let report_data = pal.dma_alloc(io, REPORT_DATA_LEN)?;
+    report_data.copy_from_slice(report_data_src);
```
jaygmsft: Why do we need this alloc and copy? Looking at the code looks to me that we can use report_data_src directly.

fw/core/lib/src/ddi/mbor/hkdf_derive.rs
```diff
@@ -48,8 +48,10 @@ pub(crate) async fn hkdf_derive<'p, P: HsmPal>(
 
     let algo = super::from_ddi::hash(body.hash_algorithm)?;
     let target = super::kdf::resolve_target(body.key_type, body.key_length)?;
+    // A KDF-derived key is created on-device, so `local = true` (it attests
```
jaygmsft, line 51: Per PKCS11, only Generate can set local=true and Copy of object that already has local=true. A derive operation would mean local=false.

vsonims: Let's have a comprehensive fix across fw, sim and emu.

fw/core/lib/src/ddi/mbor/attest_key.rs
vsonims: Reused the existing shared mapping `super::from_pal::ecc_curve(kind)` (returns `HsmResult<HsmEccCurve>`) and removed the local `ecc_curve_for_kind`. The branch is now `if let Ok(curve) = super::from_pal::ecc_curve(kind)`, so non-ECC kinds fall through to the RSA/symmetric arms. Fixed in 41704c92.

fw/core/lib/src/ddi/mbor/attest_key.rs
vsonims: Reused `super::from_pal::rsa_key(kind)` plus `HsmRsaKey::modulus_len()` and removed the local `rsa_modulus_len_for_kind`. The branch is now `else if let Ok(rsa) = super::from_pal::rsa_key(kind) { let mod_len = rsa.modulus_len(); ... }`. Fixed in 41704c92.

fw/core/lib/src/ddi/mbor/attest_key.rs
vsonims: Removed. There is no longer an intermediate binding — the request field is passed straight into `KeyReportParams { report_data: body.report_data, .. }`, where the `&mut DmaBuf` reborrows to `&DmaBuf` at the struct field. Fixed in 41704c92.

fw/core/lib/src/ddi/mbor/attest_key.rs
vsonims: Good catch — dropped the alloc + copy; `report_data` now borrows directly from the decoded request buffer (`report_data: body.report_data`). `key_report` calls `params.validate()` first thing in both passes, which already enforces the exact `REPORT_DATA_LEN`, so the manual length check was redundant and was removed as well. Fixed in 41704c92.

fw/core/lib/src/ddi/mbor/attest_key.rs
```diff
@@ -0,0 +1,207 @@
...
+        pal,
+        io,
+        HsmSessId::from(sess_id),
+    )?);
+
+    let vm_launch_id = crate::part_state::part_vm_launch_guid(pal, io)?;
+
+    // Derive the attested key's public component in big-endian COSE form.
+    let priv_key = pal.vault_key(io, key_id)?;
+    let key = if let Ok(curve) = super::from_pal::ecc_curve(kind) {
```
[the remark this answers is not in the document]
vsonims: Good call — moved `let priv_key = pal.vault_key(io, key_id)?;` into the ECC and RSA arms, so the symmetric/bulk/HMAC/secret path (`AttestedPubKey::Symmetric`) no longer fetches the private key material. Key existence is still validated up front via `vault_key_kind`/`vault_key_attrs`, so a bad `key_id` still errors early; we just avoid touching secret bytes when there is no public component to derive. Fixed in 3251ac45.

jaygmsft (approved):

PR #568, opened 2026-07-16
feat(fw): implement ReopenSession handler + live-migration session re…

jaygmsft:
…nego

Wire the ReopenSession DDI command and the live-migration session lifecycle for the emulator (std PAL). The firmware core is shared with real hardware; only the PAL differs.

- Add the reopen_session MBOR handler (new reopen_session.rs): it authenticates the credential and derives BK_SESSION via the same helpers as OpenSession, then unmasks the host-persisted bmk_session to recover MK_SESSION and re-keys the migrated session slot. A malformed bmk_session surfaces as MaskedKeyDecodeFailed, consistent with the other masked-key unmask paths.
- Route ReopenSession as an in-session command: it reuses the caller's existing session id, so it surfaces no new CQE session id. The central session-liveness gate added in #561 (io.rs validate_session_live) already returns SessionNeedsRenegotiation for in-session ops on a migrated slot and exempts ReopenSession, so no per-op renego check is needed in dispatch.
- std PAL: preserve live sessions across the NSSR as NeedsRenegotiation (SessionTable::backup/restore in part_disable_internal) and skip the old-key cleanup in session_create for renego slots -- mirrors the reference firmware's migrate() = restore(backup()) + recreate_session. A migrated slot is reopen-only: a fresh open never reclaims it, so a fully-migrated table reports VaultSessionLimitReached.
- ReopenSession reports PartitionNotProvisioned before the credential check, since it re-supplies its own credential in the request.
- emu dev: erase() preserves the file-handle session so the in-session reopen clears the host-side session check; DdiEmuDev closes its session on Drop (MBOR + TBOR) so the process-global emulator partition does not leak the now-preserved sessions across tests and exhaust the table. TBOR session-id tracking records the firmware-assigned id from the SessionOpenInit response (the TBOR analog of MBOR's hdr.sess_id) and clears it on SessionClose, mirroring exec_op_mbor.

Tests: add DDI smoke coverage for reopen, live migration (renego, plus a full-table migrate that rejects fresh opens with VaultSessionLimitReached while reopen still succeeds), the open-session limit, and close-on-drop freeing a slot; add SessionTable unit tests for the renego reclamation rules and a TBOR CU-range-exhaustion integration test.

emu DDI suite (azihsm_ddi_mbor_types --features emu): 556 -> 581 passing of 585. The 4 remaining failures (get_cert_chain x2, masked_key_get_unwrapping_with_lm,
rsa_unwrap_generated_key::test_rsa_unwrap_aes_key) are pre-existing on main and unrelated to sessions. emu TBOR (azihsm_ddi_tbor_types --features emu): 91/91. emu smoke gate: 83/83. Mock stays green; uno builds clean (uno PAL functional fix deferred to a follow-up).

Copilot-Session: 82c8bd53-6639-46fc-8a79-a7539921e4c1

### DIFF
--- a/ddi/emu/src/dev.rs
+++ b/ddi/emu/src/dev.rs
@@ -24,11 +24,15 @@ use azihsm_ddi_mbor_codec::MborDecode;
 use azihsm_ddi_mbor_codec::MborDecoder;
 use azihsm_ddi_mbor_codec::MborEncoder;
 use azihsm_ddi_mbor_types::DdiAesOp;
+use azihsm_ddi_mbor_types::DdiApiRev;
+use azihsm_ddi_mbor_types::DdiCloseSessionCmdReq;
+use azihsm_ddi_mbor_types::DdiCloseSessionReq;
 use azihsm_ddi_mbor_types::DdiDecoder;
 use azihsm_ddi_mbor_types::DdiDeviceKind;
 use azihsm_ddi_mbor_types::DdiOp;
 use azihsm_ddi_mbor_types::DdiOpReq;
 use azihsm_ddi_mbor_types::DdiOpenSessionCmdResp;
+use azihsm_ddi_mbor_types::DdiReqHdr;
 use azihsm_ddi_mbor_types::DdiRespHdr;
 use azihsm_ddi_mbor_types::DdiStatus;
 use azihsm_ddi_mbor_types::MborError;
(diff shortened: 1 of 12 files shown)


jaygmsft:
@microsoft-github-policy-service rerun

radutta99 (approved):

vsonims (approved):

PR #580, opened 2026-07-18
feat(tbor): implement HmacGenerateKey + Hmac (masked-key HMAC crypto)

vsonims:
## Summary

Adds the TBOR **HMAC crypto command pair**, stacked on top of the
RSA-AES key-unwrap commands (#583).

> Reordered stack: this PR now sits **above** the key-unwrap PR (#583).
> It reuses the shared session/masking helpers and the shared `HashAlgo`
> wire enum introduced by #583, so this layer is purely the HMAC command
> surface.

### `HmacGenerateKey` (opcode `0x11`)

Generate a fresh random HMAC key of the caller-selected SHA variant
(SHA-256/384/512 → 32/48/64 B) and return it **masked** (AEAD-GCM-256)
under the requested scope's masking key. The key is not stored
on-device; the caller holds the masked blob (unmask-on-use).

### `Hmac` (opcode `0x12`)

Compute an HMAC tag over a host-supplied message using a caller-held
masked HMAC key, unmasked on-device for the operation and then
discarded.

## Wiring

Both opcodes are routed through the fw dispatcher, the `is_known_opcode`
/ `is_in_session` / `needs_session_id_cross_check` classifiers, and
`op::SessionCtrl::from_tbor_opcode` (InSession, CO/CU).

## Tests / validation

- Emu: keygen round-trip across all hashes/scopes, MAC round-trip,
  tamper-reject, scope/hash rejects.
- Full TBOR emu suite green (126 tests at this layer); `clippy` clean;
  nightly `fmt` + `copyright` clean; Uno (`thumbv7em`) build passes.
- Docs: `docs/tbor-ddi/commands/{hmac_generate_key,hmac}.md` + README
  rows (`0x11`, `0x12`).

### DIFF
--- a/ddi/tbor/types/src/hmac.rs
+++ b/ddi/tbor/types/src/hmac.rs
@@ -1,0 +1,85 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Host-side wrapper for the TBOR `Hmac` command.
+//!
+//! `Hmac` is an **in-session** command (Crypto-Officer or Crypto-User)
+//! that computes an HMAC tag over `msg` using a caller-held **masked**
+//! HMAC key (the `masked_key` returned by
+//! [`HmacGenerateKey`](crate::hmac_generate_key) or imported via unwrap).
+//! The device unmasks the key on-device, computes the MAC, and returns the
+//! tag; nothing is persisted.
+
+use alloc::vec::Vec;
+
+pub use crate::hmac_generate_key::MASKED_HMAC_KEY_MAX_LEN;
+pub use crate::hmac_generate_key::MASKED_HMAC_KEY_MIN_LEN;
+use crate::tbor;
+
+/// TBOR opcode for `Hmac`.
+pub const TBOR_OP_HMAC: u8 = 0x12;
+
+/// Maximum message length (bytes) accepted by `Hmac`.
+///
+/// `1024`, matching the MBOR `Hmac` command's message bound
+/// (`DdiHmacReq::msg`), which is itself sized to the AES-CBC
+/// single-shot message cap (`MAX_MSG_SIZE`) so the two symmetric
+/// primitives share one host-visible limit.  Larger inputs are hashed
+/// host-side or chunked by the caller.
+pub const HMAC_MSG_MAX_LEN: usize = 1024;
+
+/// Maximum HMAC tag length (bytes): the SHA-512 digest.
+pub const HMAC_TAG_MAX_LEN: usize = 64;
+
+/// Host-facing TBOR `Hmac` request.
+#[tbor(opcode = TBOR_OP_HMAC, session_ctrl = in_session)]
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct TborHmacReq {
+    /// Session id this request is bound to.  Cross-checked against the
+    /// SQE-carried session id by the dispatcher.
+    #[tbor(session_id)]
+    pub session_id: u16,
+
+    /// The masked HMAC key (from `HmacGenerateKey` / unwrap), an
+    /// AEAD-GCM-256 envelope of 164..=260 B.
+    #[tbor(min_len = 164, max_len = 260)]
+    pub masked_key: Vec<u8>,
+
+    /// The message to MAC, up to [`HMAC_MSG_MAX_LEN`] (1024) bytes.
+    #[tbor(max_len = 1024)]
+    pub msg: Vec<u8>,
+}
+
+/// Host-facing TBOR `Hmac` response.
+#[tbor(response)]
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct TborHmacResp {
+    /// The HMAC tag over `msg`: 32 / 48 / 64 B for SHA-256 / 384 / 512.
+    #[tbor(max_len = 64)]
+    pub tag: Vec<u8>,
+}
+
+#[cfg(test)]
+mod tests {
+    use azihsm_ddi_tbor_types::TborOpReq;
+
+    use super::*;
+
+    #[test]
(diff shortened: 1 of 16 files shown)


ddi/tbor/types/src/hmac.rs
```diff
@@ -0,0 +1,79 @@
...
+
+pub use crate::hmac_generate_key::MASKED_HMAC_KEY_MAX_LEN;
+pub use crate::hmac_generate_key::MASKED_HMAC_KEY_MIN_LEN;
+use crate::tbor;
+
+/// TBOR opcode for `Hmac`.
+pub const TBOR_OP_HMAC: u8 = 0x12;
+
+/// Maximum message length (bytes) accepted by `Hmac`.
+pub const HMAC_MSG_MAX_LEN: usize = 1024;
```
jaygmsft, line 29: Just 1k?

ddi/tbor/types/src/hmac_generate_key.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// SQE-carried session id by the dispatcher.
+    #[tbor(session_id)]
+    pub session_id: u16,
+
+    /// Requested key scope as the 1-byte `KeyScope` discriminant.
+    pub scope: u8,
+
+    /// Hash algorithm as the 1-byte `HashAlgo` discriminant (see
+    /// [`HMAC_HASH_SHA256`] / [`HMAC_HASH_SHA384`] / [`HMAC_HASH_SHA512`]).
+    pub hash_algo: u8,
```
jaygmsft, line 57: This is different from var hmac.

ddi/tbor/types/src/hmac.rs
vsonims: Yes — kept at 1024 to follow MBOR: the MBOR `Hmac` command caps `msg` at 1024, which is itself sized to the AES-CBC single-shot message cap (`MAX_MSG_SIZE = 1024`) so the two symmetric primitives share one host-visible limit. Documented the rationale on `HMAC_MSG_MAX_LEN` (host + fw) in 3e14f882 so it no longer reads as an arbitrary value. Larger inputs are pre-hashed or chunked host-side.

ddi/tbor/types/src/hmac_generate_key.rs
vsonims: Good catch — reworked `HmacGenerateKey` into a true variable-length HMAC keygen in f37ee8e9, following the reference firmware (`~/mcr-hsm`) `VarLenHmacSha*` model. The request now carries a required `key_length` validated to the SHA variant's `[min, max]` range (SHA-256: 32–64, SHA-384: 48–128, SHA-512: 64–128) — out-of-range (incl. 0) is rejected with `InvalidKeyLength`. The key is generated at that length and stamped `VarLenHmacSha*`; masked-key max grows to 260 B (128-byte key), and the `Hmac` command's `masked_key` input max grows to match. Added emu coverage for the min/mid/max lengths per variant and the range rejects.

fw/core/lib/src/ddi/tbor/hmac_generate_key.rs
```diff
@@ -0,0 +1,198 @@
...
+                alloc,
+                AeadAlg::AesGcm256,
+                masking_key,
+                &params,
+                key_buf,
+                Some(out.masked_key),
+            )
+            .await;
+            key_buf.zeroize();
+            mask_res.map(|_| ())
```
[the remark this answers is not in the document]
vsonims: Fixed in 8926bc59 — the handler now binds `mask`'s return value and returns `InternalError` unless it wrote exactly `masked_len` bytes, so no uninitialized trailing bytes can remain in the reserved `masked_key` slot.

docs/tbor-ddi/commands/hmac_generate_key.md
```diff
@@ -0,0 +1,93 @@
...
+## Request
+
+### TOC entries
+
+| Offset | Field | Type | Description |
+|---|---|---|---|
+| 4 | `session_id` | `session_id` (inline) | Session this request is bound to; cross-checked against the SQE-carried session id. |
+| 8 | `scope` | `uint8` (inline) | Requested key scope (`KeyScope` discriminant): `1` = Session, `2` = Ephemeral, `3` = Local, `4` = SecurityDomain. |
+| 12 | `hash_algo` | `uint8` (inline) | HMAC hash variant (`HashAlgo` discriminant): `1` = SHA-256, `2` = SHA-384, `3` = SHA-512. |
+| 13 | `key_length` | `uint8` (inline) | Requested key length in bytes; must be in the variant's `[min, max]` range (see table above). |
```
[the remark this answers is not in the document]
vsonims: Fixed in 8926bc59 — `key_length` is now listed at TOC offset 16 (4-byte aligned after session_id@4, scope@8, hash_algo@12).

docs/tbor-ddi/commands/hmac.md
```diff
@@ -0,0 +1,71 @@
...
+## Description
+
+Computes an HMAC tag over a host-supplied message using a caller-held
+**masked** HMAC key (the `masked_key` returned by
+[`HmacGenerateKey`](./hmac_generate_key.md) or imported via unwrap).
+
+The masked key's scope is read from its cleartext, tag-bound metadata to
+select the masking key; the key is unmasked on-device (verifying the AEAD
+tag), the MAC is computed, and the tag is returned.  Nothing is
+persisted, and the recovered key is wiped from the per-IO DMA pool.
```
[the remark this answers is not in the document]
vsonims: Fixed in 8926bc59 — the doc now says the recovered key is unmasked and wiped **in place in the request buffer** (`decode_mut` + `masked_key.zeroize()`), matching the handler, rather than "per-IO DMA pool".

jaygmsft (approved):

PR #582, opened 2026-07-18
feat(engine): load EC private keys from azihsm:// URIs

walterchris:
## Summary

Adds OpenSSL 1.1.x **engine** support for loading EC private keys from the HSM via `azihsm://` key URIs (`ENGINE_load_private_key`).

- **`load_private_key` hook** in the engine-core toolkit (`Engine::set_load_privkey` + `LoadPrivKeyHandler`), mirroring the existing `set_destroy` plumbing. The C trampoline returns `*mut EVP_PKEY`, so it wraps its body in `catch_panic` with a NULL fallback via a new `result_to_ptr` helper (the pointer-returning analogue of `result_to_int`).
- **EC key loading**: a key id is an `azihsm://<masked-key-file>;type=<ec|rsa|rsa-pss>` URI (parsed in `uri.rs`, mirroring the provider's STORE). Loading opens the HSM on first use, unmasks the key pair, rebuilds the public `EVP_PKEY` from the SDK's DER, and attaches the live HSM private key to the `EC_KEY` ex_data for later operations. The key is owned by `EngineData` and deleted from the HSM when the engine is destroyed — not via an ex_data free callback, which would be unsafe across `.so` unload.
- **Engine-bound key lifetime**: the returned `EC_KEY` is built with `EC_KEY_new_method`, which takes a functional reference on the engine. So the engine — and the `EngineData` that owns the loaded HSM keys — cannot be torn down while any loaded key (or an `EVP_PKEY` built from it) is alive, and the HSM-key pointer stashed in ex_data can never dangle.
- RSA/RSA-PSS URIs parse but return a clear not-supported error for now (RSA loading needs an HSM import path and lands with its own tests).

## Testing

Full engine suite passes on OpenSSL 1.1.x (Ubuntu 20.04 / 1.1.1f): build, clippy `-D warnings` (engine + `azihsm_crypto`), unit tests, doctests, and the `openssl engine -t` load check.

- **Unit** — a mock round-trip (`load_ec_key_round_trips_through_engine`) exercises the full generate → export masked blob → load via the engine → verify public key → free cycle (HSM keys are deleted at engine teardown). Key-loader logic is covered directly: `read_masked_key` round-trips a regular file and rejects a symlink (`O_NOFOLLOW`), a non-regular file, and an oversized blob; `load_key` rejects RSA/RSA-PSS before opening the HSM.
- **Integration** — two suites drive the real `ENGINE_load_private_key` path against a masked blob staged out-of-process, both wired into the engine matrix CI:
  - **CLI** (`openssl pkey -engine azihsm -inform engine -in "azihsm://…;type=ec" -pubout`).
  - **ABI** (`ENGINE_by_id("dynamic")` + `SO_PATH`/`ID`/`LOAD` → `ENGINE_load_private_key`).

## Hardware tests

Two `#[ignore]`d smoke tests live in `context.rs` `hw_tests` (compiled only in non-mock builds, so they never run in the mock CI cell). They validate a real device end to end — e.g. the TPM OBK/POTA sources the mock can't exercise:

- `open_from_env_smoke` — HSM open only (partition open → init → session).
- `load_ec_key_from_env_smoke` — the full key-loading round trip against a real device.

Run them on a provisioned HSM host after configuring the `AZIHSM_*` environment:

```bash
export AZIHSM_CREDENTIALS_ID=<32 hex>  AZIHSM_CREDENTIALS_PIN=<32 hex>
export AZIHSM_RESILIENCY_ENABLED=1                     # storage dir + MOBK/POTA persistence
export AZIHSM_OBK_SOURCE=tpm  AZIHSM_POTA_SOURCE=tpm   # source selection
# Storage dir must exist, be mode 0700, and be owned by you (override with
# AZIHSM_RESILIENCY_STORAGE_DIR); default is /var/lib/azihsm/resiliency:
sudo install -d -m 700 -o "$USER" /var/lib/azihsm/resiliency
umask 0077

cargo test -p azihsm_ossl_engine --features engine \
  open_from_env_smoke load_ec_key_from_env_smoke -- --ignored --nocapture
```


plugins/ossl_engine/azihsm_ossl_engine/src/context.rs
```diff
@@ -412,10 +460,19 @@ mod tests {
...
     use std::sync::atomic::Ordering;
 
+    use azihsm_api::HsmEccCurve;
+    use azihsm_api::HsmEccKeyGenAlgo;
+    use azihsm_api::HsmKeyClass;
+    use azihsm_api::HsmKeyCommonProps;
+    use azihsm_api::HsmKeyKind;
+    use azihsm_api::HsmKeyManager;
+    use azihsm_api::HsmKeyPropsBuilder;
+    use foreign_types::ForeignType;
```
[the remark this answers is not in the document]
walterchris: Only compiles on Linux through mod context guard

plugins/ossl_engine/azihsm_ossl_engine/src/context.rs
```diff
@@ -571,6 +694,17 @@ mod tests {
 /// ```
 #[cfg(all(test, not(feature = "mock")))]
```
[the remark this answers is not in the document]
walterchris: Only compiles on Linux through mod context guard

plugins/ossl_engine/azihsm_ossl_engine/src/keyload.rs
```diff
@@ -0,0 +1,226 @@
...
+    // Hand the key to EngineData, which owns it and deletes it from the HSM when
+    // the engine is destroyed (while this `.so` is loaded). Stash the non-owning
+    // pointer it returns for later sign/derive retrieval. The ex_data slot has
+    // no free callback (see ec_key_ex_index), so libcrypto never calls back into
+    // this `.so` for cleanup — even if the EVP_PKEY outlives the engine.
+    let key_ptr = data.retain_loaded_key(key);
+    // SAFETY: ec is valid, idx is a registered EC_KEY ex_data slot, and key_ptr
+    // is a non-owning pointer into a key EngineData keeps alive for the engine's
+    // lifetime; the slot has no free callback, so no ownership is transferred.
+    let rc = unsafe { ffi::EC_KEY_set_ex_data(ec, idx, key_ptr.cast_mut().cast::<c_void>()) };
```
mhatrevi, line 247: The `EC_KEY` stores a non-owning pointer into `EngineData.loaded_keys`, but `EngineData` is dropped when the engine is destroyed even if this `EVP_PKEY` is still alive. In that ordering, the ex_data pointer becomes dangling. It is not currently dereferenced, but the planned sign/derive callbacks would turn this into a use-after-free.

plugins/ossl_engine/azihsm_ossl_engine/src/context.rs
```diff
@@ -494,6 +587,72 @@ mod tests {
...
+                    priv_key.pub_key_der_vec().unwrap(),
+                ))
+            })
+            .unwrap();
+
+        let blob_path = scratch.0.join("ec_key.bin");
+        write_key_material(&blob_path, &masked).unwrap();
+
+        let uri = format!("azihsm://{};type=ec", blob_path.display());
+        let raw = crate::keyload::load_key(&data, &uri).unwrap();
```
mhatrevi: This round-trip calls `crate::keyload::load_key` directly, so it does not exercise the newly added `ENGINE_load_private_key` integration: callback registration, the C trampoline, engine ex_data lookup, `CStr` conversion, error-queue handling, or ownership transfer through OpenSSL.

Could we keep this direct unit test and add an integration test that loads the key through `ENGINE_load_private_key` on a configured engine? That would validate the primary behavior introduced by this PR end to end.

mhatrevi (dismissed):

plugins/ossl_engine/integration-tests/openssl-capi/src/engine_capi_tests.rs
```diff
@@ -0,0 +1,182 @@
...
+        assert!(!ec.is_null(), "loaded EVP_PKEY has no EC_KEY");
+        assert!(
+            !ffi::EC_KEY_get0_public_key(ec).is_null(),
+            "loaded key carries no public point"
+        );
+        ffi::EVP_PKEY_free(raw);
+    }
+
+    let _ = std::fs::remove_dir_all(&dir);
+}
```
rajesh-gali, line 182: missing tests to cover keyload.rs

Cargo.toml
```diff
@@ -69,6 +69,9 @@ members = [
   "plugins/ossl_engine/azihsm_ossl_engine_core",
   "plugins/ossl_engine/azihsm_ossl_engine_resiliency",
   "plugins/ossl_engine/azihsm_ossl_engine_sys",
+  "plugins/ossl_engine/integration-tests/masked-keygen",
+  "plugins/ossl_engine/integration-tests/openssl-capi",
+  "plugins/ossl_engine/integration-tests/openssl-cli",
```
rajesh-gali, line 74: can we leverage tests from provider? it would be better if we have single test source for both engine and provider, and have conditional checks for unsupported.

plugins/ossl_engine/azihsm_ossl_engine/src/uri.rs
```diff
@@ -0,0 +1,174 @@
...
+    let key_type = key_type
+        .ok_or_else(|| EngineError::Other("azihsm URI missing required type= attribute".into()))?;
+
+    Ok(KeyUri {
+        masked_key_path: PathBuf::from(path),
+        key_type,
+    })
+}
+
+fn parse_key_type(value: &str) -> EngineResult<KeyType> {
```
rajesh-gali: missing function comments. Also, function name suggests intention is to parse the key type, not string. I think we should use fromstr to make it more clear

plugins/ossl_engine/azihsm_ossl_engine/src/lib.rs
```diff
@@ -26,6 +26,57 @@ pub mod context;
...
+/// Integration-test helper: open the HSM from the ambient `AZIHSM_*` environment,
+/// generate a persistent EC P-384 key, and return its masked blob — the form the
+/// loader consumes. Reuses the engine's real open path so a generated blob and a
+/// later `ENGINE_load_private_key` share the same masking (via the persisted BMK
+/// under the shared resiliency storage dir).
+///
+/// Gated behind `integration`; it is not part of the production engine and the
+/// cdylib OpenSSL loads is built without this feature.
+#[cfg(all(target_os = "linux", feature = "engine", feature = "integration"))]
+pub fn generate_masked_ec_p384_from_env() -> azihsm_ossl_engine_core::error::EngineResult<Vec<u8>> {
```
rajesh-gali: I think this is not the right place for this function, because of two reasons. 1. This is a test helper function 2. integration feature only. I would recommend to move this function to integration module

plugins/ossl_engine/azihsm_ossl_engine/src/lib.rs
```diff
@@ -26,6 +26,57 @@ pub mod context;
...
+            HsmKeyManager::generate_key_pair(session, &mut algo, priv_props, pub_props)
+                .map_err(|e| EngineError::wrap("generate EC key pair", e))?;
+        priv_key
+            .masked_key_vec()
+            .map_err(|e| EngineError::wrap("export masked key", e))
+    })
+}
+
 #[cfg(all(target_os = "linux", feature = "engine"))]
 mod engine_impl {
```
rajesh-gali: I would recommend extract engine_impl to seperate module file, following existing pattern

plugins/ossl_engine/azihsm_ossl_engine/src/context.rs
```diff
@@ -403,6 +474,53 @@ fn hex_decode_16(s: &str, var: &'static str) -> EngineResult<[u8; 16]> {
...
+#[allow(unsafe_code)]
+#[allow(clippy::unwrap_used)]
+fn new_test_engine() -> (
+    azihsm_ossl_engine_core::engine::Engine,
+    *mut azihsm_ossl_engine_core::ffi::ENGINE,
+) {
+    use std::ptr::NonNull;
+
+    use azihsm_ossl_engine_core::engine::Engine;
+    use azihsm_ossl_engine_core::ffi;
```
rajesh-gali: [NIT] use statements within the function

Cargo.toml
walterchris: Engine and Provider have fundamentally different CAPIs and cli usage. I would not know where to unify those..

Or do you mean that we use C++/gtest for the engine integration tests (even though we can't reuse the actual tests?)

plugins/ossl_engine/azihsm_ossl_engine/src/context.rs
walterchris: write_key_material and new_test_engine are shared by both mod tests
  ▎ (#[cfg(feature = "mock")]) and mod hw_tests (#[cfg(not(feature = "mock"))]), so they need to stay at module scope rather than move into either test module.
  
  I moved the use statements up to a module-scope.

rajesh-gali (approved):

PR #584, opened 2026-07-18
feat(tbor): implement AesGenerateKey + AesEncryptDecrypt (masked-key AES)

vsonims:
## Summary

Adds the TBOR **AES crypto command pair**, stacked on the RSA-AES
key-unwrap commands (#583), following the MBOR AES commands but with the
TBOR masked-key (stateless) model — no vault `key_id` / `key_tag`.

### `AesGenerateKey` (opcode `0x15`)

Generate a fresh random AES key (128 / 192 / 256 bits, non-bulk only,
mirroring MBOR) and return it **masked** (AEAD-GCM-256) under the
requested scope's masking key. The masked-blob length is fixed by the key
size, so the handler reserves the response slot up front and masks the
generated key straight into it (encoder `*_reserve` + `decode_mut`) — no
scratch buffer, no copy.

### `AesEncryptDecrypt` (opcode `0x16`)

AES-**CBC** encrypt or decrypt a host-supplied message with a caller-held
masked AES key. **Fully zero-copy:**
- request `masked_key` is `#[tbor(mutable)]` → the handler `decode_mut`s
  the request and `unmask`s the blob **in place**, using the recovered
  key directly (no blob / key scratch copy);
- the response `msg` / `iv` slots are **reserved**, and the non-in-place
  `aes_cbc_enc_dec` reads the request message and writes the ciphertext /
  chaining IV **straight into the response** (no message copy).

Rejects non-AES key kinds (`InvalidKeyType`) and direction-permission
mismatches (`InvalidPermissions`); the recovered key is wiped on every
path (including the unmask-failure path).

### AES key import

Already provided by **`UnwrapKey`** (0x14, key class `Aes`) in #583 — an
emu test exercises the `UnwrapKey` → `AesEncryptDecrypt` round-trip.

## Wiring

Both opcodes are routed through the fw dispatcher, the `is_known_opcode`
/ `is_in_session` / `needs_session_id_cross_check` classifiers, and
`op::SessionCtrl::from_tbor_opcode` (InSession, CO/CU). Reuses the shared
`validate_active_session` / `resolve_masking_key` helpers and the encoder
reserve/fill support from #583.

## Tests / validation

- Emu: keygen round-trip across all sizes/scopes, before-finalize +
  SD-scope + unknown-size rejects; encrypt→decrypt round-trip all sizes,
  CBC chaining IV, tamper + bad-length rejects, and the `UnwrapKey`→AES
  cross-command round-trip.
- Full TBOR emu suite green (**127 tests**), `clippy` clean, nightly
  `fmt` + `copyright` clean, Uno (`thumbv7em`) build passes.
- Docs: `docs/tbor-ddi/commands/{aes_generate_key,aes_encrypt_decrypt}.md`
  + README rows (`0x15`, `0x16`).


ddi/tbor/types/src/aes_encrypt_decrypt.rs
```diff
@@ -0,0 +1,100 @@
...
+/// Host-facing TBOR `AesEncryptDecrypt` response.
+#[tbor(response)]
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct TborAesEncryptDecryptResp {
+    /// The transformed message (same length as the input `msg`).
+    #[tbor(max_len = 1024)]
+    pub msg: Vec<u8>,
+
+    /// The updated chaining IV (the last ciphertext block).
+    #[tbor(min_len = 16, max_len = 16)]
```
jaygmsft: I think we have Len=, for exact length

jaygmsft (dismissed):

ddi/tbor/types/src/aes_encrypt_decrypt.rs
vsonims: Good call — done in c0822157. FW schema now uses the exact-length attribute `#[tbor(buffer, len = 16)]` on both IV fields (request + response) instead of `min_len = 16, max_len = 16`. On the host side `#[tbor(len = N)]` is rejected as redundant, so the IVs are now typed as the fixed array `[u8; 16]` (the host's exact-length idiom). Also rebased onto latest main (#583 merged).

ddi/tbor/types/tests/commands/aes_generate_key.rs
```diff
@@ -0,0 +1,157 @@
...
+fn aes_generate_key_session_scope_before_finalize_emu() {
+    // Session-scoped keys are masked under the per-session masking key, so
+    // they do not require a finalized partition — only an Active session.
+    let ctx = TestCtx::new();
+    let session = bootstrap_rotated_co(&ctx, &ROTATED_CO_PSK);
+    roundtrip(&ctx, session.session_id, SCOPE_SESSION, AES_KEY_SIZE_256);
+}
+
+#[test]
+fn aes_generate_key_rejects_security_domain_scope_emu() {
```
jaygmsft, line 134: the tests should have _emu in their name if they should work on both emu and real hardware eventually.

ddi/tbor/types/tests/commands/aes_generate_key.rs
```diff
@@ -0,0 +1,157 @@
...
+//! * Happy path per key size (128/192/256) — the masked key has the
+//!   expected length (148/156/164 B) and is non-zero; a second call yields
+//!   a distinct key.
+//! * Every masking-key scope: `Session` (works pre-finalize), `Ephemeral`
+//!   / `Local` (provisioned by `PartFinal`).
+//! * `SecurityDomain` scope before `CreateSD` → `UnsupportedKeyScope`.
+//! * `Ephemeral` scope before `PartFinal` → `InvalidArg`.
+//! * Unknown key size → `InvalidArg`.
+
+#![cfg(feature = "emu")]
```
jaygmsft: We shouldnt limit the tbor types to just emu

jaygmsft (dismissed):
approving with suggestions

ddi/tbor/types/tests/commands/aes_generate_key.rs
vsonims: Done in `2f3f0f2b` — moved both AES type-test modules off the file-level `#![cfg(feature = "emu")]` to the `api_rev.rs` pilot pattern.

The module gate is now `#![cfg(any(feature = "emu", feature = "mock", feature = "sock"))]`, so the shared masked-key helpers/constants (`generate_key`, `SCOPE_*`, `aes_op`, `key_len_for_size`) are available under any backend rather than being emu-limited. Only the FW-handler-dependent imports and the individual tests carry `#[cfg(feature = "emu")]`.

Verified the module compiles cleanly under `--features emu` / `mock` / `sock` and with no backend feature.

ddi/tbor/types/tests/commands/aes_generate_key.rs
vsonims: Done in `2f3f0f2b` — the emu-only tests keep their `_emu` suffix and now each carries an explicit `#[cfg(feature = "emu")]` (moved off the blanket file gate).

So the `_emu` name + per-test gate marks exactly the tests that need the emu FW handler today, while the module itself is gated on `any(emu, mock, sock)` and can host `sock` / real-hardware tests later without touching the shared helpers.

fw/core/lib/src/ddi/tbor/aes_generate_key.rs
```diff
@@ -0,0 +1,164 @@
...
+                AeadAlg::AesGcm256,
+                masking_key,
+                &params,
+                key_buf,
+                Some(out.masked_key),
+            )
+            .await;
+            key_buf.zeroize();
+            mask_res.map(|_| ())
+        })
```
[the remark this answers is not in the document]
vsonims: Fixed in `86337b4a` — the handler now binds `mask`'s return value and enforces `written == masked_len`, returning `InternalError` otherwise (same guard as `unwrap_key` / `hmac_generate_key`). The raw key is still zeroized on every path before the length check propagates, so a short write can no longer leave uninitialized trailing bytes in the reserved `masked_key` slot.

fw/core/lib/src/ddi/tbor/aes_encrypt_decrypt.rs
```diff
@@ -0,0 +1,183 @@
...
+//!
+//! Within an open session, AES-CBC encrypt or decrypt a host-supplied
+//! message using a caller-held **masked** AES key (the `masked_key` from
+//! [`AesGenerateKey`](super::aes_generate_key) or imported via
+//! [`UnwrapKey`](super::unwrap_key)).  The masked key's scope is read from
+//! its cleartext, tag-bound metadata to select the masking key; the key is
+//! unmasked on-device (verifying the AEAD tag), the AES-CBC transform runs
+//! in place, and the transformed message plus updated chaining IV are
+//! returned — nothing is persisted and the recovered key is wiped.  This
+//! is the TBOR analogue of MBOR `AesEncryptDecrypt`, keyed by a masked blob
```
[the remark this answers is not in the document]
vsonims: Fixed in `86337b4a` — corrected the module doc. The **key** is unmasked in place in the request buffer (`decode_mut` + `unmask`), but the AES-CBC **transform** is zero-copy *out-of-place*: `aes_cbc_enc_dec` reads `req.msg` and writes into the reserved `resp.msg` / `resp.iv` slots. Reworded to "the AES-CBC transform runs zero-copy — reading the request message and writing the transformed message plus updated chaining IV straight into the response buffer".

docs/tbor-ddi/commands/aes_encrypt_decrypt.md
```diff
@@ -0,0 +1,82 @@
...
+
+AES-**CBC** encrypts or decrypts a host-supplied message using a
+caller-held **masked** AES key (the `masked_key` from
+[`AesGenerateKey`](./aes_generate_key.md) or imported via
+[`UnwrapKey`](./unwrap_key.md)). The device reads the masked key's scope
+from its cleartext, tag-bound metadata to select the masking key, unmasks
+the key on-device (verifying the AEAD tag), runs the AES-CBC transform in
+place, and returns the transformed message plus the updated chaining IV so
+the host can chain subsequent CBC blocks. This is the TBOR analogue of
+MBOR `AesEncryptDecrypt`, keyed by a masked blob rather than a vault
```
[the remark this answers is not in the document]
vsonims: Fixed in `86337b4a` — updated the command doc to match the implementation: the transform is zero-copy but out-of-place (reads the request message, writes the result straight into the response buffer), not in-place. Replaced the "runs the AES-CBC transform in place" wording accordingly.

fw/core/lib/src/ddi/tbor/aes_generate_key.rs
```diff
@@ -0,0 +1,173 @@
...
+    {
+        let out = TborAesGenerateKeyResp::decode_mut(resp)?;
+        pal.alloc_scoped_async(io, async |alloc| -> HsmResult<()> {
+            // Generate the random AES key into scratch.
+            let key_buf = alloc.dma_alloc(key_len)?;
+            pal.aes_gen_key(io, key_buf).await?;
+
+            let masking_key = resolve_masking_key(pal, io, scope, sess_id)?;
+            let key_label = alloc.dma_alloc(AES_KEY_LABEL.len())?;
+            key_label.copy_from_slice(AES_KEY_LABEL);
```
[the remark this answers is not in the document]
vsonims: Fixed in `ff12a1cf` — reordered so the raw key can never outlive an early return.

`resolve_masking_key(...)` and the `key_label` allocation (the two fallible steps that were between keygen and the zeroize) now run **before** `pal.aes_gen_key(...)`, so an unsupported/unprovisioned scope or a label-alloc failure is rejected before any key material exists — which also makes the module doc's "an unavailable scope fails cheaply before any key is generated" actually true (previously `SecurityDomain` → `UnsupportedKeyScope` resolved *after* keygen).

Keygen + mask now run inside an inner block whose `Result` is bound; `key_buf.zeroize()` runs unconditionally after it, so the key is wiped on **every** post-generation path — keygen failure, mask failure, or the short-write guard — before the error propagates.

All 13 AES emu tests still pass (incl. `aes_generate_key_rejects_security_domain_scope_emu`, which now rejects pre-keygen).

jaygmsft (approved):

PR #585, opened 2026-07-18
feat(tbor): implement ECC keygen, sign, and ECDH derive

vsonims:
## Summary

Adds three stateless **masked-key** TBOR crypto DDI commands, mirroring the existing MBOR vault-key-handle equivalents. The private key is AEAD-GCM-256 **masked** and returned to the host (nothing persisted on-device); the caller passes the masked blob back on use.

| Opcode | Command | Handler |
|---|---|---|
| `0x17` | `EccGenerateKey` | generate keypair → masked private key + wire public key |
| `0x18` | `EccSign` | unmask masked key in place → ECDSA `r‖s` over a host digest |
| `0x19` | `EcdhDerive` | unmask local key in place → ECDH vs. peer pub → re-mask secret |

ECC **import** is already provided by the existing `UnwrapKey` (Ecc class) from the base PR; a cross-command test covers `UnwrapKey`-Ecc → `EccSign`.

## Design (mirrors the sibling TBOR handlers)
- **Zero-copy**: keygen masks the private key straight into a reserved response slot; sign/ecdh use `decode_mut` to **unmask the masked key in place** (no scratch copy) and reserve+fill outputs.
- **Secret hygiene**: recovered plaintext keys/secrets are scrubbed on **every** return path (scope exit only resets the bump watermark — it does not zero freed DMA memory).
- **Endianness**: device emits PKA-native wire-LE; host-side test verification reverses per-coordinate (confirmed by passing OpenSSL ECDSA verification for all curves).
- **Strict key-class handling**: `EccSign`/`EcdhDerive` recover the curve via a strict `from_pal::ecc_private_curve` that rejects non-ECC-private kinds (incl. `SdSealing`); usage attrs (`sign`/`derive`) enforced post-unmask.

## Changes
- Handlers: `fw/core/lib/src/ddi/tbor/{ecc_generate_key,ecc_sign,ecdh_derive}.rs`.
- Shared PAL mappings added to `fw/core/lib/src/ddi/tbor/from_pal.rs` (`ecc_private`, `ecdh_secret`, `ecc_private_curve`).
- Wiring: `mod.rs` (opcodes, dispatch, 3 classifiers), `op.rs` (`from_tbor_opcode`).
- Wire schemas (fw + host mirror) + `lib.rs` registration.
- Emu integration tests (13): keygen all curves/scopes/reject; sign roundtrip **host-verified** for all curves + `UnwrapKey`-Ecc→sign cross-test + rejects; ecdh all curves/scopes + reject.
- Docs: 3 command `.md` + README rows.

## Testing
- `cargo check` fw core (host + **Uno** no_std) ✅
- clippy (fw + host) ✅ · nightly fmt ✅ · copyright ✅
- Full TBOR emu suite: **127 passed** (114 baseline + 13 ECC).

## Notes
- Based on #583 (`tbor/unwrap-key`); uses opcodes `0x17`–`0x19` to avoid collision with the sibling AES/HMAC branches.
- Reviewed with the code-review and security-review specialists: two secret-scrubbing gaps on error paths were found and fixed; security review found no exploitable vulnerabilities.

### DIFF
--- a/ddi/tbor/types/src/ecc_sign.rs
+++ b/ddi/tbor/types/src/ecc_sign.rs
@@ -1,0 +1,74 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Host-side wrapper for the TBOR `EccSign` command.
+//!
+//! `EccSign` is an **in-session** command (Crypto-Officer or Crypto-User)
+//! that produces a raw ECDSA `r ‖ s` signature over a host-supplied
+//! **pre-computed digest** using a caller-held **masked** ECC private key.
+//! Firmware does no hashing — the caller supplies the digest in wire
+//! little-endian order.
+
+use alloc::vec::Vec;
+
+use crate::tbor;
+
+/// TBOR opcode for `EccSign`.
+pub const TBOR_OP_ECC_SIGN: u8 = 0x18;
+
+/// Maximum digest length (bytes): the SHA-512 digest.
+pub const ECC_DIGEST_MAX_LEN: usize = 64;
+/// Maximum wire ECDSA signature length (`r ‖ s`, P-521 padded).
+pub const ECC_SIG_MAX_LEN: usize = 136;
+
+/// Host-facing TBOR `EccSign` request.
+#[tbor(opcode = TBOR_OP_ECC_SIGN, session_ctrl = in_session)]
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct TborEccSignReq {
+    /// Session id this request is bound to.
+    #[tbor(session_id)]
+    pub session_id: u16,
+
+    /// The masked ECC private key (from `EccGenerateKey` / `UnwrapKey`).
+    #[tbor(min_len = 164, max_len = 200)]
+    pub masked_key: Vec<u8>,
+
+    /// The pre-computed message digest in wire little-endian order. Its
+    /// length must be one of the supported SHA-2 digest lengths
+    /// (32 / 48 / 64 B); the hash algorithm is inferred from that length,
+    /// so no separate algorithm selector is carried on the wire.
+    #[tbor(max_len = 64)]
+    pub digest: Vec<u8>,
+}
+
+/// Host-facing TBOR `EccSign` response.
+#[tbor(response)]
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct TborEccSignResp {
+    /// Raw ECDSA `r ‖ s`, each component little-endian and padded to the
+    /// curve wire coordinate length (64 / 96 / 136 B).
+    #[tbor(max_len = 136)]
+    pub signature: Vec<u8>,
+}
+
+#[cfg(test)]
+mod tests {
+    use azihsm_ddi_tbor_types::TborOpReq;
+
+    use super::*;
+
+    #[test]
+    fn request_encodes_digest() {
+        let req = TborEccSignReq {
+            session_id: 7,
+            masked_key: alloc::vec![0x11u8; 164],
+            digest: alloc::vec![0x22u8; 32],
+        };
+        let mut buf = [0u8; 512];
+        let frame = req.encode_request(&mut buf).expect("encode");
+        assert!(
+            frame.windows(4).any(|w| w == [0x22u8; 4]),
+            "encoded frame must carry the digest bytes",
+        );
+    }
+}
--- a/fw/core/lib/src/ddi/tbor/ecc_generate_key.rs
+++ b/fw/core/lib/src/ddi/tbor/ecc_generate_key.rs
@@ -1,0 +1,182 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! TBOR `EccGenerateKey` command handler.
+//!
+//! Within an open session, generate a fresh ECC keypair on the requested
+//! NIST curve, mask the private key (AEAD-GCM-256) under the requested
+//! scope's masking key, and return the masked blob plus the wire public
+//! key.  The private key is **not** persisted on-device: the caller holds
+//! the masked blob and passes it back to [`EccSign`](super::ecc_sign) /
+//! [`EcdhDerive`](super::ecdh_derive) (unmask-on-use).  This is the TBOR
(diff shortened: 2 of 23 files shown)


fw/core/lib/src/ddi/tbor/ecc_sign.rs
```diff
@@ -0,0 +1,112 @@
...
+            let frame = TborEccSignResp::encode(buf, 0, false)?
+                .signature_reserve(curve.wire_sig_len())?
+                .finish();
+            Ok(frame.as_bytes().len())
+        })?;
+        {
+            let out = TborEccSignResp::decode_mut(resp)?;
+            pal.ecc_sign(io, curve, view.target_key, req.digest, out.signature)
+                .await?;
+        }
```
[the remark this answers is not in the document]
vsonims: Fixed in 71a1bbae — `EccSign` no longer passes the raw host digest to `pal.ecc_sign`. It now copies the digest into a zeroed **operand-width** buffer (`curve.wire_coord_len()` = 32 / 48 / 68) and signs `&operand[..curve.ecdsa_digest_len()]`, so the PKA's full-operand DMA read (68 B for P-521) stays in bounds over zero pad, mirroring the MBOR handler. Digests longer than `ecdsa_digest_len` are rejected with `InvalidArg` (added an emu test for SHA-512 + P-256).

fw/core/lib/src/ddi/tbor/ecc_generate_key.rs
```diff
@@ -0,0 +1,176 @@
...
+                    io,
+                    alloc,
+                    AeadAlg::AesGcm256,
+                    masking_key,
+                    &params,
+                    priv_key,
+                    Some(out.masked_key),
+                )
+                .await
+                .map(|_| ())
```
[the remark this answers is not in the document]
vsonims: Fixed in 71a1bbae — the handler now binds `mask`'s return and returns `InternalError` unless it wrote exactly `masked_len`, so no uninitialized trailing bytes can remain in the reserved `masked_key` slot (same pattern as `unwrap_key`).

fw/core/lib/src/ddi/tbor/ecdh_derive.rs
```diff
@@ -0,0 +1,183 @@
...
+                        alloc,
+                        AeadAlg::AesGcm256,
+                        masking_key_target,
+                        &params,
+                        secret,
+                        Some(out.masked_secret),
+                    )
+                    .await
+                    .map(|_| ())
+                }
```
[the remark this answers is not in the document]
vsonims: Fixed in 71a1bbae — the derived-secret path now validates `written == masked_len` after `mask` and returns `InternalError` otherwise, so a short write cannot leak uninitialized DMA bytes in `masked_secret`.

ddi/tbor/types/src/ecc_sign.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[tbor(session_id)]
+    pub session_id: u16,
+
+    /// The masked ECC private key (from `EccGenerateKey` / `UnwrapKey`).
+    #[tbor(min_len = 164, max_len = 200)]
+    pub masked_key: Vec<u8>,
+
+    /// Hash algorithm of the supplied digest, 1-byte `HashAlgo` (see
+    /// [`ECC_DIGEST_SHA256`] / [`ECC_DIGEST_SHA384`] / [`ECC_DIGEST_SHA512`]).
+    pub digest_algo: u8,
```
jaygmsft: I think we can infer digest algo from digest length unless shake is in question.

fw/core/lib/src/ddi/tbor/ecc_generate_key.rs
```diff
@@ -0,0 +1,182 @@
...
+    req_buf: &DmaBuf,
+) -> HsmResult<&'p DmaBuf> {
+    let req = TborEccGenerateKeyReq::decode(req_buf)?;
+    let sess_id = HsmSessId::from(u16::from(req.session_id()));
+    let scope = HsmKeyScope(req.scope().0);
+    validate_active_session(pal, io, sess_id)?;
+
+    // Non-`Session` scopes need a masking key provisioned by `PartFinal` /
+    // the SD lifecycle; fail cheaply before keygen.
+    if scope != HsmKeyScope::Session && part_state::part_state(pal, io)? != PartState::Initialized {
```
jaygmsft: Is this check correct? Other FSMs like ecc sign dont seem to have it.

ddi/tbor/types/src/ecc_sign.rs
vsonims: Done in `d56870ef` — removed the `digest_algo` field and now infer the hash from the digest length.

In the TBOR design the host sends the digest at its **exact** length, so `req.digest.len()` already *is* the digest length and the field was pure redundancy. (MBOR needs `digest_algo` because its host zero-pads the digest to the 68-byte PKA operand width, so firmware uses the algo to locate the real digest inside the padding — TBOR doesn't do that.)

The handler now rejects any length that isn't a supported SHA-2 digest length (32/48/64); SHAKE/XOF isn't in scope here. Updated fw + host schemas, handler, tests (added `ecc_sign_unsupported_digest_len_rejected_emu`), and the command doc.

fw/core/lib/src/ddi/tbor/ecc_generate_key.rs
vsonims: Good catch — removed it in `d56870ef` (from both `ecc_generate_key` and `ecdh_derive`).

The check was redundant with, and coarser than, the real gate: `resolve_masking_key` already validates each scope's masking key at mask time and returns the documented `UnsupportedKeyScope` when a scope isn't provisioned yet. `EccSign` / `UnwrapKey` rely on exactly that resolution, so dropping the pre-check makes all the masked-key handlers consistent and yields the more-specific error. The old blanket `part_state == Initialized` check was also too coarse (it gated every non-`Session` scope on partition finalization rather than checking the specific scope's masking key).

Updated the `InvalidArg` rows in the `ecc_generate_key` / `ecdh_derive` docs accordingly (the unprovisioned-scope case is already documented as `UnsupportedKeyScope`).

jaygmsft (approved):

PR #589, opened 2026-07-19
feat(tbor): implement HKDF and single-step (X9.63 / SP 800-56A) KDF derivation

vsonims:
## Summary

Adds two in-session TBOR **KDF** commands that derive key material from a caller-held **masked ECDH shared secret** (from [`EcdhDerive`](https://github.com/Azure/azihsm-sdk/pull/585) `0x19`) and return the derived key **masked** under the requested scope — the KDF analogues of the ECC crypto commands, re-masking the output instead of vaulting it.

| Opcode | Command | KDF |
|---|---|---|
| `0x1C` | `HkdfDerive` | HKDF (RFC 5869), salt + info |
| `0x1D` | `ConcatKdfDerive` | single-step: ANSI **X9.63** or NIST **SP 800-56A** one-step (selected by `kdf_alg`), single info |

Both derive AES (128/192/256) or HMAC (fixed SHA-256/384/512, or variable-length) keys.

## Design
- **Unmask-on-use**: the ECDH secret is unmasked **in place** in the request buffer, checked to be an ECDH-secret kind carrying `derive`, run through the KDF into scratch, and the output masked under the target scope. The recovered secret and derived scratch are **scrubbed on every path**.
- **Shared `KdfKeyType`** output-type enum lives in `key_props` (used by both commands).
- **Cross-platform single-step KDF in `azihsm_crypto`**: X9.63 and SP 800-56A are Hash-only, so one generic `ConcatKdfAlgo` over the shared `Hasher` serves **both OpenSSL (Linux) and CNG (Windows)** — no per-backend code. The std PAL `x963_kdf` / `sp800_56a_kdf` hooks (previously `todo!()` stubs) now delegate to it via `StdKdf::concat_kdf`; the Uno PAL already implemented them.

## Changes
- **`azihsm_crypto`**: `ConcatKdfAlgo` + `ConcatKdfMode {X963, Sp800_56a}` (`crates/crypto/src/kdf/concat.rs`), 3 `CryptoError` variants, **known-answer tests** (`kdf/tests/concat_tests.rs`).
- **std PAL**: `StdKdf::concat_kdf` + wired both PAL methods; `HsmError::ConcatKdfError`.
- **TBOR**: wire schemas (fw + host mirror), handlers, dispatch + 3 classifiers + `op.rs` for `0x1C` / `0x1D`.
- **Tests**: emu coverage for every output key type under each KDF (and each X9.63 / SP 800-56A variant), all hashes/scopes, optional salt/info, and rejects (unknown hash / KDF variant / key type, var-HMAC missing / out-of-range length, non-ECDH-secret IKM).
- **Docs**: `hkdf_derive.md`, `concat_kdf_derive.md`, README rows.

## Testing
- `azihsm_crypto`: **501 lib tests pass** (incl. X9.63 / SP 800-56A KATs vs. independently computed vectors).
- Full TBOR emu suite: **146 pass** (19 new).
- `cargo check` fw core (host + **Uno** no_std) + Uno PAL ✅ · clippy (crypto / std-PAL / fw / host) ✅ · nightly fmt ✅ · copyright ✅.

## Notes
- Based on **#585** (`tbor/ecc-crypto`) for the ECDH shared-secret input. Uses opcodes `0x1C` / `0x1D` after the sibling AES / ECC / RSA / Hash crypto branches (`0x15..0x1B`).
- `ConcatKdfDerive` is **TBOR-only** (no MBOR analogue). The concat KDFs' byte-level correctness is proven by the `azihsm_crypto` KATs; the emu tests verify command plumbing (the derived key is only observable masked).


fw/core/lib/src/ddi/tbor/hkdf_derive.rs
```diff
@@ -0,0 +1,308 @@
...
+    let target_scope = HsmKeyScope(req.scope);
+    validate_active_session(pal, io, sess_id)?;
+
+    // The derived key is masked under `target_scope`; a non-`Session` scope
+    // needs a masking key provisioned by `PartFinal` / the SD lifecycle.
+    // Fail cheaply before any crypto work.
+    if target_scope != HsmKeyScope::Session
+        && part_state::part_state(pal, io)? != PartState::Initialized
+    {
+        return Err(HsmError::InvalidArg);
```
jaygmsft, line 196: Bunch of in session commands dont have this check, if the check is needed, should it be centralized?

jaygmsft (dismissed):

rajesh-gali (approved):

jaygmsft (approved):

PR #590, opened 2026-07-20
More open session tests

wenbo-yuan:
Add more tbor open session tests for HW (comparing to the mbor open session tests)
Clean up description section.


jaygmsft (changes requested):
Please write tests that work on both emu and hw.

PR #592, opened 2026-07-21
Merge hw tests with existing emu

wenbo-yuan:
This PR merges all the existing HW tests with emu tests. Most of the tests are working both on emu and HW. 
There are quite a few HW only tests that need emu support. This should be fixed in the follow-up PR.


ddi/tbor/types/tests/harness/mod.rs
```diff
@@ -42,12 +42,23 @@
 //! `get_certificate`) carry per-method `#[cfg(feature = "emu")]` and
 //! are unavailable under `--features mock`.
 
-#![cfg(any(feature = "emu", feature = "mock", feature = "sock"))]
+#![cfg(any(
```
jaygmsft: Given that hw also supports these tests, I think we dont need any cfg(any()) statement here.

ddi/tbor/types/tests/harness/hw_ctx.rs
```diff
@@ -0,0 +1,501 @@
...
+use crate::harness::session::session_open_init_with_options as session_open_init_with_options_helper;
+use crate::harness::session::PendingHandshake;
+use crate::harness::session::SessionHandshake;
+use crate::harness::session::SessionOpenInitOptions;
+
+/// Fixed default 48-byte SATA thumbprint used by the convenience
+/// [`HwCtx::part_init`] wrapper.
+const DEFAULT_SATA_THUMBPRINT: [u8; 48] = [0x5A; 48];
+
+/// Process-global serialisation lock. A single physical device is
```
jaygmsft: nextest has a parameter to run single threaded. why not just use that instead of having this file.

ddi/tbor/types/tests/commands/api_rev.rs
```diff
@@ -62,9 +58,9 @@ fn api_rev_repeated_stable_emu() {
 /// transitions to Active. Together with the gate test in
 /// `default_psk_gate.rs` this proves the dispatcher never lets
 /// session state leak into the out-of-session handler.
-#[cfg(feature = "emu")]
+#[cfg(not(any(feature = "mock", feature = "sock")))]
```
jaygmsft: I think we can remove the sock as well:

```suggestion
#[cfg(not(feature = "mock"))]
```

jaygmsft: Actually thinking more we can just do the mock thing once at the beginning of the file:

```
#![cfg(not(feature = "mock"))]
```

ddi/tbor/types/tests/azihsm_ddi_tbor_tests.rs
```diff
@@ -6,16 +6,10 @@
...
-//! OS backend (`nix` on Linux / `win` on Windows), which drives real
-//! silicon. Hardware-only smoke tests live under [`hw`].
+//! `--features mock` (transport-contract probes). With **no** feature
+//! enabled the crate falls through to the native OS backend (`nix`
+//! on Linux / `win` on Windows) for on-silicon test runs.
 
-#[cfg(any(feature = "emu", feature = "mock", feature = "sock"))]
 pub mod harness;
 
 pub mod commands;
```
jaygmsft, line 21: We could also disable mock by doing it here:

```suggestion
#[cfg(not(feature = "mock"))]
pub mod commands;
```

ddi/tbor/types/tests/commands/default_psk_gate.rs
```diff
@@ -1,34 +1,29 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#![cfg(not(any(feature = "mock", feature = "sock")))]
```
jaygmsft: We dont need to worry about sock.

ddi/tbor/types/tests/commands/part_info.rs
```diff
@@ -3,27 +3,30 @@
...
+//! Backend is selected at compile time by
+//! [`azihsm_ddi::AzihsmDdi::default`].
 
 use azihsm_ddi_tbor_types::TborPartInfoReq;
 
 use crate::harness::TestCtx;
 
 /// `DdiDeviceKind::Physical` discriminant — uno is a physical device.
-#[cfg(any(feature = "emu", feature = "sock"))]
+#[cfg(not(feature = "mock"))]
```
jaygmsft: I know its not your doing but it would be great if you could help make thing cfg not thing consistent. Mostly we only need to worry about not mock so that can be global or once per file instead of per method and we need not worry about sock.

ddi/tbor/types/tests/commands/psk_change.rs
```diff
@@ -46,6 +49,17 @@ use crate::harness::TestCtx;
 const CO: u8 = 0;
 const CU: u8 = 1;
 
+/// Backend-specific status for wire-length rejects.
+///
+/// Emu's decoder passes the frame through to the handler, which
+/// returns `TborInvalidFixedLength`. The hw decoder trips the
+/// schema's `#[tbor(buffer, len = 100)]` check earlier and surfaces
+/// `DdiDecodeFailed` before the handler runs.
+#[cfg(feature = "emu")]
```
jaygmsft: FW based on plat/uno will also return this error code same as emu. I think you need to ensure that code you write returns this error code.

ddi/tbor/types/tests/harness/ctx.rs
```diff
@@ -65,26 +83,131 @@ use crate::harness::session::SessionOpenInitOptions;
...
+/// The primary fd sits behind `Arc<Dev>` (not `Arc<TestDev>`)
+/// deliberately: `TestDev` bundles the `parking_lot::MutexGuard` and
+/// is therefore `!Send`; wrapping it in `Arc` would make `TestCtx`
+/// `!Sync`, which in turn breaks `std::thread::scope`-based tests
+/// that share `&TestCtx` across threads. Keeping the guard as a
+/// separate field of `TestCtx` leaves the ctx `!Send` (the guard
+/// still can't cross a `Send` boundary) but `Sync`, which is what
+/// `s.spawn(|| ctx.foo())` actually needs.
 pub struct TestCtx {
-    dev: TestDev,
```
jaygmsft, line 72: Why do we need this change? What was the issue with TestDev?

ddi/tbor/types/tests/harness/fixture.rs
```diff
@@ -66,20 +66,64 @@ impl Deref for TestDev {
...
+/// `AZIHSM_MAX_SESSIONS_PER_FD = 1`, so every concurrent session past
+/// the first sits on its own fd. On `emu` / `mock` / `sock` the extra
+/// handle is just another view onto the process-global instance —
+/// harmless.
+///
+/// The caller must already hold the process-global lock via the
+/// primary [`TestDev`]; that is what makes it safe for this function
+/// to bypass [`TEST_LOCK`]. Never call this without a live primary in
+/// scope.
+pub fn open_extra_dev() -> <AzihsmDdi as Ddi>::Dev {
```
jaygmsft: Why do we need this method? Are you trying to reopen the same device again? If so then this code is not guaranteed to do that.

ddi/tbor/types/Cargo.toml
```diff
@@ -39,11 +39,6 @@ path = "tests/azihsm_ddi_tbor_tests.rs"
 emu = ["azihsm_ddi/emu"]
 mock = ["azihsm_ddi/mock"]
 sock = ["azihsm_ddi/sock"]
```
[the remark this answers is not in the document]
wenbo-yuan: No that is the expected behavior.

ddi/tbor/types/tests/azihsm_ddi_tbor_tests.rs
```diff
@@ -6,16 +6,16 @@
...
+//! `mock` disables the whole `commands` tree — mock rejects TBOR at
+//! the transport layer, so command-level integration tests are
+//! meaningless there. Under `mock` this binary compiles the harness
+//! only and runs zero tests.
 
-#[cfg(any(feature = "emu", feature = "mock", feature = "sock"))]
 pub mod harness;
 
+#[cfg(not(feature = "mock"))]
 pub mod commands;
```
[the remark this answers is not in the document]
wenbo-yuan: This is by design, that the default implementation is on hardware.

ddi/tbor/types/tests/harness/fixture.rs
```diff
@@ -77,9 +90,32 @@ pub fn open_dev() -> TestDev {
...
+/// acquired: the primary `TestDev` already holds `TEST_LOCK`, and
+/// extras are conceptually part of the same test's device set.
+///
+/// Caller must ensure the primary `TestDev` outlives every extra
+/// `Dev`, otherwise the lock guard drops mid-test.
+pub fn open_extra_dev(path: &str) -> <AzihsmDdi as Ddi>::Dev {
+    AzihsmDdi::default()
+        .open_dev(path)
+        .expect("open extra backend device on the same path as the primary TestDev")
 }
```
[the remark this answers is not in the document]
wenbo-yuan: It is already mentioned in the comment secion that lock is not needed in this function.

ddi/tbor/types/tests/commands/default_psk_gate.rs
```diff
@@ -111,17 +105,84 @@ fn default_psk_gate_session_close_bypass_emu() {
...
+
+/// A non-allow-listed in-session opcode (`PartInit`) is rejected at
+/// dispatch with `DefaultPskMustRotate`. The FW returns the gate
+/// error before any partition-state mutation, so this is safe to run
+/// on real silicon.
+#[test]
+fn default_psk_gate_part_init_rejected() {
+    use azihsm_ddi_tbor_types::PolicyKeyKind;
+    use azihsm_ddi_tbor_types::TborStatus;
+    use azihsm_ddi_tbor_types::MACH_SEED_LEN;
```
jaygmsft: Lets have all the use statements once at the top of the file. I think some previous code had it that way because of the flags stuff.

ddi/tbor/types/tests/harness/fixture.rs
```diff
@@ -77,9 +90,32 @@ pub fn open_dev() -> TestDev {
...
+/// Used by multi-fd tests (e.g. `open_session_multiple_concurrent`)
+/// whose semantics require overlapping sessions on distinct fds — on
+/// hw the driver enforces `AZIHSM_MAX_SESSIONS_PER_FD = 1`, so the
+/// second concurrent session must live on its own handle. No lock is
+/// acquired: the primary `TestDev` already holds `TEST_LOCK`, and
+/// extras are conceptually part of the same test's device set.
+///
+/// Caller must ensure the primary `TestDev` outlives every extra
+/// `Dev`, otherwise the lock guard drops mid-test.
+pub fn open_extra_dev(path: &str) -> <AzihsmDdi as Ddi>::Dev {
```
jaygmsft: I think we should talk about a better name for this method

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +246,566 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
-    let mut guards = Vec::new();
-    for _ in 0..7 {
-        guards.push(ctx.open_session(CU, SessionType::PlainText));
+    // Each concurrent session lives on its own extra `Dev` bound to
+    // the same underlying device — required on hw where
+    // `AZIHSM_MAX_SESSIONS_PER_FD = 1`. Devs are kept alive alongside
+    // their session ids so we can issue the matching close on the
+    // same fd.
+    type Dev = <azihsm_ddi::AzihsmDdi as azihsm_ddi_interface::Ddi>::Dev;
+    let mut open_slots: Vec<(Dev, u16)> = Vec::new();
```
jaygmsft: Do we need to specify the type explicitly for open_slots? I dont think so, and in that case line 291 also goes away.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +246,566 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+    let mut rejection_seen = false;
+
+    for _ in 0..16 {
```
jaygmsft: why 16?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +246,566 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+                );
+                rejection_seen = true;
+                break;
+            }
+        }
     }
-    let err = ctx
+
+    // Close everything we opened before running the recovery check,
+    // so a recovery-side failure does not leak slots on the board.
```
jaygmsft: Do you really leak slots if you dont close explicitly? If so that is a product bug in tbor sessions and we need to fix it.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +246,566 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
-    // Keep the remaining guards live until here so the table stayed full
-    // for the rejection above.
-    drop(guards);
-    let _ = err;
+        .expect("session table must recover after close-all");
+    ctx.session_close(recovered.session_id)
+        .expect("SessionClose after recovery must succeed");
+
+    if !rejection_seen {
+        eprintln!(
```
jaygmsft: The test will still pass if you do this right? why not assert?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +246,566 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+// Emu's `Arc<StdHsm>` is a process-global singleton and its TBOR
+// handler chain is not reentrant across the two-phase handshake:
+// concurrent inits clobber shared state mid-flight and surface as
+// `SessionAuthFailure` / `SessionNotFound` at finish instead of the
+// clean table-full / winner-takes-slot behaviour real FW produces.
+// The property under test only holds on the native OS backend.
+// ---------------------------------------------------------------------------
+
+#[cfg(not(feature = "emu"))]
+const MULTI_THREADED_TOTAL: usize = 12;
```
jaygmsft: Why are these gated for not emu?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +246,566 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+                handles.push(
+                    s.spawn(|| -> Result<(Dev, u16), azihsm_ddi_interface::DdiError> {
+                        let dev = open_extra_dev(path);
+                        let pending = session_open_init_on_dev(&dev, CU, SessionType::PlainText)?;
+                        let handshake = session_open_finish_on_dev(&dev, pending)?;
+                        Ok((dev, handshake.session_id))
+                    }),
+                );
+            }
+            let mut winners: Vec<(Dev, u16)> = Vec::new();
```
jaygmsft: We shouldnt have to specify the data type here

ddi/tbor/types/tests/commands/psk_change.rs
```diff
@@ -302,3 +326,51 @@ fn psk_change_wrong_aad_length_emu() {
...
+    assert_fw_rejects(&err, TborStatus::InvalidArg);
+}
+
+#[cfg(not(feature = "emu"))]
+#[test]
+fn psk_change_cu_to_default_cu_rejected() {
+    run_psk_change_to_default_rejected(CU, SessionType::PlainText, &DEFAULT_PSK_CU);
+}
+
+#[cfg(not(feature = "emu"))]
```
jaygmsft, line 376: Why do we need to gate emu for these?

ddi/tbor/types/tests/harness/session_guard.rs
```diff
@@ -65,11 +71,11 @@ impl<'ctx> SessionGuard<'ctx> {
...
     /// Consuming `self` makes double-close a *compile* error rather
     /// than a runtime one — tests that *want* to assert the FW
-    /// rejects a double-close must drive the second
-    /// [`TestCtx::session_close`] call themselves.
+    /// rejects a double-close must drive the second `session_close`
+    /// call themselves.
     pub fn close(mut self) -> DdiResult<()> {
         self.closed = true;
-        self.ctx.session_close(self.handshake.session_id)
+        self.ctx.close_session_by_id(self.handshake.session_id)
```
jaygmsft: why do we need this change?

wenbo-yuan: I think you are looking at an outdated change. This file is no longer touched in the latest commit.

ddi/tbor/types/tests/commands/psk_change.rs
wenbo-yuan: Right now emu doesnt support this rejection. We need emu support for releasing this gate.

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: The reason is already in the above comment section.

ddi/tbor/types/tests/harness/fixture.rs
```diff
@@ -77,9 +90,32 @@ pub fn open_dev() -> TestDev {
     let ddi = AzihsmDdi::default();
     let infos = ddi.dev_info_list();
     let info = infos.first().expect("backend should advertise a device");
-    let dev = ddi.open_dev(&info.path).expect("open test backend device");
-    #[cfg(feature = "emu")]
+    let path = info.path.clone();
+    let dev = ddi.open_dev(&path).expect("open test backend device");
+    #[cfg(not(feature = "mock"))]
```
jaygmsft: do we need to guard for not mock here given we dont run tbor tests against mock at all? or is there another guard missing at higher layer somewhere?

ddi/tbor/types/tests/commands/psk_change.rs
jaygmsft: lets chat about these

ddi/tbor/types/tests/commands/default_psk_gate.rs
```diff
@@ -78,6 +79,8 @@ fn default_psk_gate_session_open_init_bypass_emu() {
     let session_co = ctx
         .session_open_finish(pending_co)
         .expect("CO finish under default PSK");
+    ctx.session_close(session_co.session_id)
```
jaygmsft, line 81: Do we have a test that does both CO and CU session in parallel on two different device handles? If not can we please add?

ddi/tbor/types/tests/commands/default_psk_gate.rs
```diff
@@ -111,17 +111,76 @@ fn default_psk_gate_session_close_bypass_emu() {
...
+/// error before any partition-state mutation, so this is safe to run
+/// on real silicon.
+#[test]
+fn default_psk_gate_part_init_rejected() {
+    // Build a 484-byte `PartPolicy` blob that passes wire decode so
+    // the request reaches the dispatcher's gate. Mirrored from
+    // `commands::part_init::known_good_part_policy` (kept inline so
+    // this file stays runnable on hw without ungating the destructive
+    // part_init/ submodules).
+    fn known_good_part_policy() -> [u8; PART_POLICY_LEN] {
```
jaygmsft: Why is this function not a shared function?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -186,13 +183,17 @@ fn session_open_finish_unknown_session_id_emu() {
         .tbor(&req)
         .expect_err("finish against unknown session_id must fail");
     assert!(
-        matches!(err, azihsm_ddi_interface::DdiError::TborStatus(_)),
-        "expected FW-side rejection, got {err:?}",
+        matches!(
+            err,
+            azihsm_ddi_interface::DdiError::TborStatus(_)
+                | azihsm_ddi_interface::DdiError::DdiStatus(_)
```
jaygmsft: Why did we have to add DdiStatus? Why would a tbor command return a ddistatus?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +246,542 @@ fn session_open_finish_seed_envelope_tampered_emu() {
 // ---------------------------------------------------------------------------
 
 #[test]
-fn open_session_multiple_concurrent_emu() {
+fn open_session_multiple_concurrent() {
     let ctx = TestCtx::new();
     let a = ctx.open_session(CU, SessionType::PlainText);
```
jaygmsft: Did we add a test where we ensure that the old behavior doesnt work?

I am seeing test changes for tests to work on real hw but looks like bugs in tbor implementation arent being flagged or fixed.

```
    let a = ctx.open_session(CU, SessionType::PlainText);
    let b = ctx.open_session(CU, SessionType::PlainText);
```

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: The  DdiStatus  arm isn't FW returning a DDI status for a TBOR command — it's the Linux kernel driver rejecting the ioctl before it reaches FW.

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: I am trying to understand what is "old behavoir" and what is the "bugs in tbor implementation" 
The reason for this change, as mentioned below, is that driver does not allow multiple sessions to be opened in a single fd. That is why we need to open another dev to have another session. Do we need to add another test to make sure opening multiple sessions on a single dev doesnt work?

ddi/tbor/types/tests/harness/fixture.rs
wenbo-yuan: The harness is not gated at all both mock and sock should be able to access this code path. Since I am enabling this code path for hw, I think the right way is to add gate for not mock not soc.

jaygmsft: Why not gate harness in ddi/tbor/types/tests/azihsm_ddi_tbor_tests.rs the way commands was done or maybe the whole thing with #! syntax? I think you can gate for mock there. sock is not needed since the code for sock is same as emu.

ddi/tbor/types/tests/commands/open_session.rs
jaygmsft: in this test previously, there were two open session calls against the same ctx. both were successful based on my reading of the test. that is incorrect right? so the question is why was it working in emu, there must be a bug. we need to ensure that behavior is same across emu and hardware.

ddi/tbor/types/tests/harness/fixture.rs
wenbo-yuan: Ok so we wanna both gate mock and sock for both harness and commands?

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: The reason why the behavior is different, is that emu doesnt restrict one session per dev. That is why it works on emu but fails on HW. This restriction comes from the driver level.

ddi/tbor/types/tests/commands/default_psk_gate.rs
```diff
@@ -111,17 +113,69 @@ fn default_psk_gate_session_close_bypass_emu() {
...
+#[test]
+fn default_psk_gate_co_and_cu_parallel_on_separate_devs_bypass() {
+    let ctx = TestCtx::new();
+
+    // Session A: CO + Authenticated on the ctx's fd, under default CO PSK.
+    let session_co = ctx.open_session(CO, SessionType::Authenticated);
+
+    // Session B: CU + PlainText on a second fd bound to the same
+    // device, under default CU PSK. Held concurrently with A.
+    let dev_b = open_dev_with_path(ctx.path());
```
jaygmsft: Looks like we created a method open_dev_with_path independently. Wondering if it should have been a method inside TestCtx so you could have done TestCtx::new_with_path()? Thoughts?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -33,6 +19,13 @@ use azihsm_ddi_tbor_types::SEED_ENVELOPE_LEN;
 use azihsm_ddi_tbor_types::SESSION_SUITE_P384_HKDF_SHA384_AES_GCM_256;
 
 use crate::harness::build_mac_fin;
+use crate::harness::open_dev_with_path;
+use crate::harness::session::open_session as open_session_on_dev;
```
jaygmsft: every file seems to be renaming these methods, why not name them appropriately in the original file?

ddi/tbor/types/tests/commands/open_session.rs
jaygmsft: I dont think that should happen. Can you please guide me exactly where it is happening?

ddi/tbor/types/tests/commands/part_init/mod.rs
```diff
@@ -53,67 +59,6 @@ pub(crate) const ROTATED_CO_PSK: [u8; PSK_LEN] = [
     0xB1, 0xB2, 0xB3, 0xB4, 0xB5, 0xB6, 0xB7, 0xB8, 0xB9, 0xBA, 0xBB, 0xBC, 0xBD, 0xBE, 0xBF, 0xC0,
 ];
 
-/// Build a 484-byte unified `PartPolicy` blob that passes
-/// `azihsm_fw_hsm_core::ddi::tbor::policy::from_bytes`.  Layout mirrors
-/// the canonical wire format defined in
-/// `fw/core/ddi/tbor/types/src/policy.rs`: POTA + SATA trust anchors are
-/// populated Ecc384 keys; SAPOTA + backing-partition keys are left
-/// absent (zero `len`); flags are clear; `info` is filled.
-pub(crate) fn known_good_part_policy() -> [u8; PART_POLICY_LEN] {
```
jaygmsft, line 62: We seem to be removing these from here and moving to harness, is there a reason why we cannot refer to these methods from here?

ddi/tbor/types/tests/commands/default_psk_gate.rs
wenbo-yuan: I thought based on our last conversation, we only want single dev in the Ctx. New dev should be opened in helper function just like the existing mbor tests.

ddi/tbor/types/tests/commands/part_init/mod.rs
wenbo-yuan: I believe at this point, the part_init is still behind the emu. My next PR will open the gate but for this commit, I cant release the gate. 
Either I can bring in more changes (part init on hw), or keep this as it is. 
I dont see the issue to move this portion into the harness, as this is part of the helper functions. These are pure byte-array builders — no backend, no session state — so they belong in  harness/ . Moved to  harness/part_policy.rs  and re-exported through  harness::mod.

ddi/tbor/types/tests/commands/psk_change.rs
wenbo-yuan: Discussed about this offline. Emu support will be provided to release this gate.

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: For this test, the session id is invalid. Session id is part of the McrCpGenericCmd and when this command sends to the driver, it will be rejected directly by the driver before reaching to the FW, due to the session id mismatch. Then you will find the error mapping [here](https://github.com/Azure/azihsm-sdk/blob/96893fc2b7ac9c1dfa865b77c2a4c48b7674a568/ddi/nix/src/dev.rs#L665).

ddi/tbor/types/tests/commands/part_init/mod.rs
wenbo-yuan: Let me bring in the part init as well, since it is mainline supported.

ddi/tbor/types/tests/commands/default_psk_gate.rs
wenbo-yuan: So this is just a way to creating the TestCtx, with a given path.

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: we need a negative test to make sure this fails.

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: We need to return Tbor error mapping in the dev.exec_op_tbor.

ddi/tbor/types/tests/commands/part_init/happy_path.rs
```diff
@@ -124,6 +127,9 @@ fn part_init_smoke_roundtrip_emu() {
 /// embedded COSE_Key payload to the PTA pubkey carried in
 /// `pta_spki_der`.
 ///
+/// Emu-only: pulls in `azihsm_ddi_mbor_sim`'s attestation/COSE
```
jaygmsft: That is pulled only for host side right? What's the reason for this not working for emu? Does the HW not send the report right now or do you send in different format?

ddi/tbor/types/tests/commands/part_init/mod.rs
```diff
@@ -90,6 +92,10 @@ pub(crate) fn known_good_part_policy() -> [u8; PART_POLICY_LEN] {
 /// Like [`known_good_part_policy`] but with a caller-supplied **real**
 /// `POTAPubKey` (raw P-384 `X ‖ Y`, 96 bytes), so `PartFinal` can validate
 /// a PTA certificate chain anchored to it.
+///
+/// Emu-only: the only consumers (`part_final`, `sd_sealing_key_gen`,
+/// `sd_create_remote_backup`) are gated behind `feature = "emu"`.
```
jaygmsft: Are these not implemented for HW right now?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +244,567 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+    // secondary `TestCtx`s are kept alive in the vec so their fds
+    // stay open while we probe capacity; dropping the vec later
+    // closes every fd in one shot, which the kernel driver must
+    // translate into per-session slot reclaim on the FW side (that
+    // reclaim path is exactly what the recovery assertion below
+    // exercises).
+    let mut open_slots = Vec::new();
+    let mut rejection_seen = false;
+
+    for _ in 0..SESSION_STRESS_TOTAL {
```
jaygmsft: Constant is named SESSION_STRESS_TOTAL but this is not a stress test. Device is not under stress.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +244,567 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+                );
+                rejection_seen = true;
+                break;
+            }
+        }
+    }
+
+    let slot_count = open_slots.len();
+    assert!(
+        rejection_seen,
```
jaygmsft: The value of the constant SESSION_STRESS_TOTAL is set to 12, we have a much lower number of sessions allowed. Why not check against actual allowed session count or count the number of rejections and assert on that. This test will pass if 11 sessions open and 1 fails, when the limit of active sessions is much lower.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +244,567 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+    // Recovery: drop every extra fd in one shot so the driver reclaims
+    // their FW-side session slots, then one fresh open on the primary
+    // ctx dev must succeed. Relying on fd-drop (rather than an
+    // explicit close loop) is deliberate — the invariant we care
+    // about is that fd-drop == slot-reclaim; if that isn't true, it's
+    // a product bug worth surfacing.
+    drop(open_slots);
+
+    let recovered = ctx
+        .open_session_raw(CU, SessionType::PlainText)
```
jaygmsft: Sometimes we use open_session and sometimes we seem to use open_session_raw, whats the difference?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +244,567 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
-    // succeeds again — the rejection was a capacity limit, not a
-    // permanent failure.
-    drop(guards.pop());
-    let _reopened = ctx.open_session(CU, SessionType::PlainText);
-
-    // Keep the remaining guards live until here so the table stayed full
-    // for the rejection above.
-    drop(guards);
-    let _ = err;
+        .expect("reopen after close must succeed");
```
jaygmsft: This is not reopen but opening new session after previous session close. Reopen has a very specific meaning in the LM context.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +244,567 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+    type Dev = <azihsm_ddi::AzihsmDdi as azihsm_ddi_interface::Ddi>::Dev;
+
+    // Each entry holds the owning Dev + the session id so we can
+    // close on the correct fd once the race resolves.
+    let (winners, rejections) = std::thread::scope(|s| {
+        let mut handles = Vec::with_capacity(SESSION_STRESS_TOTAL);
+        for _ in 0..SESSION_STRESS_TOTAL {
+            handles.push(
+                s.spawn(|| -> Result<(Dev, u16), azihsm_ddi_interface::DdiError> {
+                    let dev = open_dev_with_path(path);
```
jaygmsft: Why arent we using testctx here? This helper method open_dev_with_path shouldnt be exposed to any one.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +244,567 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+        (winners, rejections)
+    });
+
+    let mut sorted_ids: Vec<u16> = winners.iter().map(|(_, sid)| *sid).collect();
+    let winner_ids = sorted_ids.clone();
+    sorted_ids.sort_unstable();
+    sorted_ids.dedup();
+    let unique_wins = sorted_ids.len();
+
+    // Close winners on their owning fds before asserting so a failing
```
jaygmsft: This will not be a problem if you use testctx iirc, right?

ddi/tbor/types/tests/commands/part_init/mod.rs
wenbo-yuan: Not yet.

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: The session table is stressed out with full entries.

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: I think this was from the existing code where it just wants to hit the limit. A re-write of this code (and also maybe other tests) would be check exactly how many can be successfully opened. Also in that case we need to split CO (1) with CU (7)

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +244,567 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+        winner_ids.len(),
+        "concurrent winners must have distinct session ids: {winner_ids:?}",
+    );
+    for err in &rejections {
+        assert!(
+            matches!(err, azihsm_ddi_interface::DdiError::TborStatus(_)),
+            "concurrent open_session rejections must be FW/driver rejections, got {err:?}",
+        );
+    }
+    if rejections.is_empty() {
```
jaygmsft: This test is just plain wrong, the title says all_should_open, and you are not asserting on rejections.is_empty(), further since SESSION_STRESS_TOTAL is 12 we know that all will never succeed so I am confused about this test.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +244,567 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+        );
+        return;
+    }
+
+    // Phase 2: free exactly one slot (close and drop the tail filler's Dev).
+    let (freed_dev, freed_id) = fillers.pop().expect("at least one filler must exist");
+    session_close_on_dev(&freed_dev, freed_id).expect("close of freed filler slot must succeed");
+    drop(freed_dev);
+
+    // Phase 3: race SESSION_STRESS_TOTAL threads for the one slot.
```
jaygmsft: This test is also incorrect, when we originally had single_winner test it was to ensure that when multiple requests go in parallel even if there are multiple session slots available, it will open a single session but here you are leaving 1 session slot and then trying to open multiple. It doesnt serve the purpose original single_winner tests had on mbor side. Please review the tests.

ddi/tbor/types/tests/harness/fixture.rs
```diff
@@ -77,9 +92,57 @@ pub fn open_dev() -> TestDev {
...
+/// Used by multi-fd tests (e.g. `open_session_multiple_concurrent`)
+/// whose semantics require overlapping sessions on distinct fds — on
+/// hw the driver enforces `AZIHSM_MAX_SESSIONS_PER_FD = 1`, so the
+/// second concurrent session must live on its own handle. No lock is
+/// acquired: the primary `TestDev` already holds `TEST_LOCK`, and
+/// extras are conceptually part of the same test's device set.
+///
+/// Caller must ensure the primary `TestDev` outlives every extra
+/// `Dev`, otherwise the lock guard drops mid-test.
+pub fn open_dev_with_path(path: &str) -> <AzihsmDdi as Ddi>::Dev {
```
jaygmsft: These shouldnt be needed given we have TestCtx::new_with_path etc.

ddi/tbor/types/tests/azihsm_ddi_tbor_tests.rs
```diff
@@ -4,18 +4,18 @@
...
-//! OS backend (`nix` on Linux / `win` on Windows), which drives real
-//! silicon. Hardware-only smoke tests live under [`hw`].
+//! `mock` and `sock` disable the whole harness + `commands` tree:
+//! mock rejects TBOR at the transport layer, and sock's `erase` is a
+//! stub, so command-level integration tests are meaningless there.
+//! Under `mock` or `sock` this binary compiles the crate only and
+//! runs zero tests.
 
-#[cfg(any(feature = "emu", feature = "mock", feature = "sock"))]
+#[cfg(not(any(feature = "mock", feature = "sock")))]
```
jaygmsft: We dont need to filter sock, only mock should be filtered.

ddi/tbor/types/tests/azihsm_ddi_tbor_tests.rs
```diff
@@ -4,18 +4,18 @@
...
 pub mod harness;
 
+#[cfg(not(any(feature = "mock", feature = "sock")))]
```
jaygmsft: same here, sock need not be filtered.

ddi/tbor/types/tests/commands/part_init/mod.rs
jaygmsft: Then why not let them fail on HW or gate with appropriate comment or TODO which makes it clear that this gate will be removed when HW support comes in. I am starting to worry that we will miss bunch of these places so we need to make sure we are tracking them with tasks etc.

ddi/tbor/types/tests/commands/open_session.rs
jaygmsft: reliability != stress.

ddi/tbor/types/tests/commands/open_session.rs
jaygmsft: We should make sure if only N sessions are allowed then only N can be open.

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: Open session returns the session guard which will drop automatically. open_session_raw returns the dev. Most of the time open session is enough. However, for this specific test, open_slots keep a copy of either dev or guard. Guard borrowed that ctx which will be dropped at the end of the loop. So it is better for a copy of dev.

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: Will rename it.

ddi/tbor/types/tests/commands/part_init/mod.rs
wenbo-yuan: I am confused about this worry. As we discussed all the existing emu tests will be released to hardware whenever the support comes. At the end of the day there will be no gate (either for emu or not emu). I am not sure how we can miss them.

ddi/tbor/types/tests/commands/open_session.rs
jaygmsft: I dont think it will be an issue since you push into the vector so the drop will happen when vector is cleared or vector is dropped not at the loop.

ddi/tbor/types/tests/commands/part_init/happy_path.rs
wenbo-yuan: It seems like this test works on the hardware. Remove the gate.

wenbo-yuan: It seems like this test works on the hardware. Remove the gate.

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: I think it has something to do with the MutexGuard being !Send. Let me see how to fix it.

ddi/tbor/types/tests/harness/fixture.rs
```diff
@@ -55,10 +55,36 @@ pub struct TestDev {
...
+// unit `()` and is never dereferenced — the guard exists purely to
+// serialise tests. Secondary `TestDev` handles (returned by
+// `open_dev_secondary`) always have `_guard: None`. Sending the
+// primary is safe as long as `TestDev` is not moved off the thread
+// that acquired the lock until the guard drops; in practice only
+// secondaries (with `_guard: None`) are sent into `thread::scope`
+// worker threads, and rustc cannot prove that at compile time
+// because both variants share the same type.
+#[allow(unsafe_code)]
+unsafe impl Send for TestDev {}
```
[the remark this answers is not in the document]
wenbo-yuan: this is a test-only harness invariant, not a product-code memory-safety issue. The  _guard  field is  Option<MutexGuard<'static, ()>> ; the lock protects  ()  and is never dereferenced.  open_dev_secondary  (the only public constructor of  TestCtx::new_with_path ) is the sole path that produces  _guard: None , and it is the only constructor invoked inside  thread::scope  worker closures — the primary  TestCtx::new()  handle always stays on the test thread that acquired the lock.
Modelling this in the type system would require splitting  TestDev  into  TestDev  (Send, no guard) and  TestDevGuarded  ( !Send ) plus a parallel  SecondaryTestCtx , and duplicating every method across both. For a ~30-line test-only invariant that's guarded by a single constructor call site, the  unsafe impl Send  + SAFETY comment is the pragmatic trade-off. Happy to revisit if the harness grows more constructors.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +237,576 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+        "expected FW-side TborStatus rejection, got {err:?}",
+    );
+    drop(overflow_ctx);
 
-    // Keep the remaining guards live until here so the table stayed full
-    // for the rejection above.
+    // Recovery: drop every guard (each guard's `Drop` sends an
+    // explicit SessionClose), then drop the extra fds. A fresh open
+    // on the primary ctx must succeed.
     drop(guards);
```
jaygmsft: dropping just one should be sufficient.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +237,576 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+
+    let ctx_second = TestCtx::new_with_path(ctx.path());
+    let err = ctx_second
+        .open_session(CO, SessionType::Authenticated)
+        .expect_err("second concurrent CO open must be rejected");
+    assert!(
+        matches!(err, azihsm_ddi_interface::DdiError::TborStatus(_)),
+        "expected FW-side TborStatus rejection, got {err:?}",
+    );
+    drop(ctx_second);
```
jaygmsft: shouldnt need to drop second.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +237,576 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+
+    // Each entry holds the owning ctx + the session id so we can
+    // close on the correct fd once the race resolves.
+    let (winners, rejections) = std::thread::scope(|s| {
+        let mut handles = Vec::with_capacity(CU_SESSION_LIMIT);
+        for _ in 0..CU_SESSION_LIMIT {
+            handles.push(s.spawn(
+                || -> Result<(TestCtx, u16), azihsm_ddi_interface::DdiError> {
+                    let ctx_extra = TestCtx::new_with_path(path);
+                    let handshake = ctx_extra.open_session_raw(CU, SessionType::PlainText)?;
```
jaygmsft: why not open_session instead of raw?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +237,576 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+    let mut sorted_ids: Vec<_> = winners.iter().map(|(_, sid)| *sid).collect();
+    let winner_ids = sorted_ids.clone();
+    sorted_ids.sort_unstable();
+    sorted_ids.dedup();
+    let unique_wins = sorted_ids.len();
+
+    // Drop winners' ctxs before asserting so a failing assert never
+    // leaves the session table dirty — each ctx's fd close reclaims
+    // its slot in FW.
+    drop(winners);
```
jaygmsft: Is this true? Because if you open session properly then it should clean up.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +237,576 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+}
+
+/// Fill CU capacity to one free slot then race `CU_SESSION_LIMIT`
+/// threads for it. Regression for FW's undo-on-loser path: every
+/// losing racer must see a clean rejection and leave the session
+/// table intact for retry. Each filler + racer holds its own
+/// [`TestCtx`] fd — hw requires one fd per concurrent session.
+#[cfg(not(feature = "emu"))]
+#[test]
+fn open_session_multi_threaded_single_winner() {
```
jaygmsft: This is not a single winner test, I had left that comment earlier as well, but cannot find it here.

ddi/tbor/types/tests/harness/fixture.rs
```diff
@@ -77,9 +105,39 @@ pub fn open_dev() -> TestDev {
...
+/// the device state, and wiping it mid-test would break both handles.
+///
+/// Used by multi-fd tests whose semantics require overlapping
+/// sessions on distinct fds — on hw the driver enforces
+/// `AZIHSM_MAX_SESSIONS_PER_FD = 1`, so the second concurrent session
+/// must live on its own handle.
+///
+/// Caller must ensure the primary [`TestDev`] outlives every
+/// secondary, otherwise the lock guard drops mid-test.
+pub(crate) fn open_dev_secondary(path: &str) -> TestDev {
```
jaygmsft, line 134: Is this still used or needed?

ddi/tbor/types/tests/harness/ctx.rs
```diff
@@ -226,15 +255,14 @@ impl TestCtx {
         pending: PendingHandshake,
         mac_fin: [u8; 48],
     ) -> DdiResult<SessionHandshake> {
-        session_open_finish_with_mac_helper(&self.dev, pending, mac_fin)
+        session_open_finish_with_mac(&self.dev, pending, mac_fin)
```
jaygmsft: I am not sure of the reason of the rename but the new names are not consistent. For example, you have session_open_finish_on_dev and session_open_finish_with_mac. 
Old names were session_open_finish_helper and session_open_finish_with_mac_helper. Werent the older names clearer?

ddi/tbor/types/tests/harness/fixture.rs
wenbo-yuan: Yes this is still used. This is used during the new_with_path. I limited to the (crate) so that outside wont see it.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +237,576 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+// ---------------------------------------------------------------------------
+// CO session-slot exhaustion + recovery
+//
+// CO has a single slot. Opening the first CO session must succeed,
+// a second concurrent CO open must be rejected, and dropping the
+// first fd must restore capacity.
+// ---------------------------------------------------------------------------
+
+#[test]
+fn open_session_fills_co_slot_then_recovers() {
```
rajesh-gali, line 368: I think we should have a test to exahust both CO + CU limit in single test to make sure whole table is full.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +237,576 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
-    // permanent failure.
-    drop(guards.pop());
-    let _reopened = ctx.open_session(CU, SessionType::PlainText);
+    // One more CU open must be rejected — table is full.
+    let overflow_ctx = TestCtx::new_with_path(ctx.path());
+    let err = overflow_ctx
+        .open_session(CU, SessionType::PlainText)
+        .expect_err("open_session past CU limit must be rejected");
+    assert!(
+        matches!(err, azihsm_ddi_interface::DdiError::TborStatus(_)),
```
rajesh-gali: Checking for blanket error could mask a different issue. Can we check for specific error like SessionLimitReached?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +237,576 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+    let first = ctx_first
+        .open_session(CO, SessionType::Authenticated)
+        .expect("first CO session must succeed");
+
+    let ctx_second = TestCtx::new_with_path(ctx.path());
+    let err = ctx_second
+        .open_session(CO, SessionType::Authenticated)
+        .expect_err("second concurrent CO open must be rejected");
+    assert!(
+        matches!(err, azihsm_ddi_interface::DdiError::TborStatus(_)),
```
rajesh-gali: Checking for blanket error could mask a different issue. Can we check for specific error like SessionLimitReached?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +237,576 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+         {rejection_count} rejections: {rejections:?})",
+    );
+    assert_eq!(
+        rejection_count,
+        racer_count - 1,
+        "all non-winning racers must fail cleanly",
+    );
+    for err in &rejections {
+        assert!(
+            matches!(err, azihsm_ddi_interface::DdiError::TborStatus(_)),
```
rajesh-gali: Checking for blanket error could mask a different issue. Can we check for specific error like SessionLimitReached?

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +237,576 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+    let _a = ctx
+        .open_session(CU, SessionType::PlainText)
+        .expect("first open on a fresh fd must succeed");
     let err = ctx
-        .open_session_raw(CU, SessionType::PlainText)
-        .expect_err("a CU handshake with the range full must be rejected");
+        .open_session(CU, SessionType::PlainText)
+        .expect_err("second open on the same fd must be rejected");
+    assert!(
+        matches!(err, azihsm_ddi_interface::DdiError::TborStatus(_)),
```
rajesh-gali: Checking for blanket error could mask a different issue. Can we check for specific error like SessionLimitReached?

ddi/tbor/types/tests/commands/part_init/happy_path.rs
```diff
@@ -245,11 +247,11 @@ fn run_part_init_capture_pta_pub(
...
 
     let bootstrap = ctx
-        .open_session_raw(CO, SessionType::Authenticated)
+        .open_session(CO, SessionType::Authenticated)
         .expect("open CO default");
-    ctx.psk_change(&bootstrap, &ROTATED_CO_PSK)
+    ctx.psk_change(bootstrap.handshake(), &ROTATED_CO_PSK)
         .expect("rotate CO PSK");
-    let _ = ctx.session_close(bootstrap.session_id);
+    let _ = bootstrap.close();
```
rajesh-gali: ignores error? and we could use bootstrap_rotated_co

ddi/tbor/types/tests/commands/part_init/happy_path.rs
```diff
@@ -307,3 +309,42 @@ fn part_init_determinism_emu() {
...
+/// `include_fmc_cdi` gates whether `fmc_cdi` is mixed into that
+/// context — and, independently, the flag byte is part of the
+/// `PartPolicy` blob whose hash is bound into the derivation — the
+/// two PTA pubkeys must not collide. A byte-identical result would
+/// mean either the flag is a stealth no-op in the KBKDF chain or
+/// the `PartPolicy` hash isn't being bound at all.
+#[test]
+fn part_init_pta_pub_differs_when_include_fmc_cdi_flag_toggled() {
+    let ctx = TestCtx::new();
+    ctx.erase().expect("erase before run 1");
```
rajesh-gali: I think ctx.erase() is redundant here.

ddi/tbor/types/tests/harness/session/session_close.rs
```diff
@@ -15,8 +15,15 @@ use azihsm_ddi_interface::DdiError;
...
 use azihsm_ddi_tbor_types::TborSessionCloseResp;
 
-/// Issue `SessionClose(session_id)` and return on success.
-pub fn session_close(dev: &<AzihsmDdi as Ddi>::Dev, session_id: u16) -> Result<(), DdiError> {
+/// Issue `SessionClose(session_id)` against a specific `dev`.
+///
+/// Named with an `_on_dev` suffix to disambiguate from the
+/// [`TestCtx::session_close`](crate::harness::TestCtx::session_close)
+/// method which acts on the ctx's implicit fd.
+pub fn session_close_on_dev(
```
rajesh-gali: I see that session_close_on_dev called from ctx::session_close only, we should be using pub(crate)

ddi/tbor/types/tests/harness/mod.rs
```diff
@@ -70,13 +60,13 @@ pub use session::build_mac_fin;
 pub use session::build_part_init_mach_seed_aad;
 pub use session::encrypt_mach_seed_envelope;
 pub use session::encrypt_psk_envelope;
-pub use session::open_session;
+pub use session::open_session_on_dev;
 pub use session::part_init;
 pub use session::psk_change;
-pub use session::session_close;
-pub use session::session_open_finish;
+pub use session::session_close_on_dev;
```
rajesh-gali: dead re-export

ddi/tbor/types/tests/harness/session.rs
```diff
@@ -41,15 +41,21 @@ pub use part_init::encrypt_mach_seed_envelope;
...
-/// `dev`. Equivalent to `session_open_init(...)? → session_open_finish(...)`.
-pub fn open_session(
+/// a specific `dev`. Equivalent to
+/// `session_open_init_on_dev(...)? → session_open_finish_on_dev(...)`.
+///
+/// Named with an `_on_dev` suffix to disambiguate from the
+/// [`TestCtx::open_session`](crate::harness::TestCtx::open_session)
+/// method which acts on the ctx's implicit fd and returns a
+/// [`SessionGuard`](crate::harness::SessionGuard).
+pub fn open_session_on_dev(
```
rajesh-gali: Not consistent, as session open is here and session close is part of session/session_close. Please move both open and close to single mod under session.

rajesh-gali: also lot of unused re-exports in the file.

ddi/tbor/types/tests/commands/psk_change.rs
```diff
@@ -302,3 +340,53 @@ fn psk_change_wrong_aad_length_emu() {
...
+// defeating the point of rotation. FW must reject with
+// `TborStatus::InvalidArg` regardless of which role requests it and
+// which default value is targeted (a CU must not be able to push the
+// CO PSK back to default either).
+//
+// Hw-only: the reject lives in the silicon firmware, not in the
+// in-tree Rust FW that emu runs. Emu accepts the rotation.
+// ===========================================================================
+
+#[cfg(not(feature = "emu"))]
```
rajesh-gali, line 358: please open a task to track changes on emu side. We should remove this emu specific gate.

ddi/tbor/types/tests/commands/open_session.rs
wenbo-yuan: Not sure if we have both CO exahustion and CU exahustion, why we need to have another test to run them together. The exahustion is based on the role.

ddi/tbor/types/tests/commands/part_init/happy_path.rs
```diff
@@ -281,7 +274,7 @@ fn run_part_init_capture_pta_pub(
 /// non-deterministic ECDSA nonces, but the PTA public key is the
 /// canonical determinism invariant under test.
 #[test]
-fn part_init_determinism_emu() {
+fn part_init_determinism() {
```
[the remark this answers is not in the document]
jaygmsft: dont worry about sock, do erase as needed, Wenbo.

ddi/tbor/types/tests/commands/open_session.rs
jaygmsft: To make sure there is no such bug.

ddi/tbor/types/tests/commands/open_session.rs
```diff
@@ -245,39 +232,480 @@ fn session_open_finish_seed_envelope_tampered_emu() {
...
+// ---------------------------------------------------------------------------
+
+/// `pk_hsm` (the partition identity pub key surfaced on Phase-1
+/// `PendingHandshake`) must be stable across handshakes — it identifies
+/// the partition, not the ephemeral session. Two sequential Phase-1s
+/// must observe byte-identical `pk_hsm`; a rotation between handshakes
+/// would break long-term binding assumed by higher-layer protocols.
+#[test]
+fn partition_pk_hsm_stable_across_handshakes() {
+    let ctx = TestCtx::new();
```
jaygmsft, line 549: Shouldnt most new tests call erase before starting the test?

ddi/tbor/types/tests/harness/session/mod.rs
```diff
@@ -1,55 +1,35 @@
...
+pub(crate) use init::PendingHandshake;
+pub(crate) use init::SessionOpenInitOptions;
+pub(crate) use open_close::session_close;
+pub(crate) use open_close::session_open;
+pub(crate) use part_final::part_final;
+pub(crate) use part_init::build_part_init_mach_seed_aad;
+pub(crate) use part_init::encrypt_mach_seed_envelope;
+pub(crate) use part_init::part_init;
+pub(crate) use psk_change::encrypt_psk_envelope;
+pub(crate) use psk_change::psk_change;
```
rajesh-gali, line 35: this is a module file, I would rename it to ddi/tbor/types/tests/harness/session/mod.rs

ddi/tbor/types/tests/harness/ctx.rs
```diff
@@ -79,10 +81,38 @@ impl TestCtx {
...
+    ///
+    /// Multi-fd tests need distinct `Dev` handles (hw enforces
+    /// `AZIHSM_MAX_SESSIONS_PER_FD = 1`); each secondary `TestCtx`
+    /// owns one such handle and otherwise behaves identically to a
+    /// primary — `open_session`, `tbor`, `mbor`, etc. all work the
+    /// same. Caller must ensure the primary `TestCtx` outlives every
+    /// secondary, otherwise the lock guard drops mid-test.
+    pub fn new_with_path(path: &str) -> Self {
+        Self {
+            dev: open_dev_secondary(path),
```
rajesh-gali, line 100: open_dev_secondary panics on failed to open dev. Should we make this try_new_With_path to give more control to the caller for the tests that need to assert/handle open failures explicitely?

jaygmsft (dismissed):

ddi/tbor/types/tests/harness/ctx.rs
wenbo-yuan: Hmm not sure what the caller will do if the dev is failed to open. it is the same thing as the primary dev, we just crash/panic.

wenbo-yuan: We will have a follow up PR that replace the panic with error which will be propagated to the caller.

wenbo-yuan: https://github.com/Azure/azihsm-sdk/issues/617

rajesh-gali (approved):

PR #599, opened 2026-07-23
feat(engine): ECDSA signing through loaded HSM keys

walterchris:
## Summary

Makes an engine-loaded EC key actually usable for signing. #582 loads the key (public `EVP_PKEY` + the live HSM key parked in `EC_KEY` ex_data) but leaves sign/derive unwired; this PR wires **ECDSA signing** through the HSM.

- **Toolkit (`azihsm_ossl_engine_core`, new `ec_method` module)** — add `EcdsaSignHandler` + a monomorphized `sign_sig` C trampoline + `new_ecdsa_method()`, which builds an `EC_KEY_METHOD` from `EC_KEY_OpenSSL` and overrides only `sign_sig` (keeping the default `sign`/`sign_setup` that DER-wrap it). Mirrors the existing `DestroyHandler`/`LoadPrivKeyHandler` pattern. A caller-supplied precomputed nonce (`kinv`/`r`) is rejected for HSM keys rather than silently ignored — the HSM generates its own nonce.
- **Loader (`azihsm_ossl_engine`)** — the engine registers that method **engine-wide** at bind time (`ENGINE_set_EC`), so every key the loader creates with `EC_KEY_new_method` adopts it at creation. The `sign_sig` hook recovers the HSM key from the `EC_KEY` ex_data, signs the caller's digest with raw ECDSA (`HsmSigner::sign_vec`), and returns `r‖s` as an `ECDSA_SIG`.
- **FFI (`azihsm_ossl_engine_sys`)** — bind `ECDSA_SIG_*` + `BN_bin2bn`/`BN_free` (only the types were bound before); include `<openssl/ecdsa.h>` + `<openssl/bn.h>`.

## Why engine-wide, not per key

Two behaviors of OpenSSL 1.1.x make the method registration load-bearing, both caught by the new integration tests:

- **`EC_KEY_set_method` drops the key's engine reference** (`ENGINE_finish(key->engine)` + `key->engine = NULL`). Attaching the method per key would let the engine — and the `EngineData` owning the loaded HSM keys — be destroyed the moment the application releases its own `ENGINE` handle, deleting the HSM keys under a still-alive `EVP_PKEY`. Registering the method engine-wide preserves the reference `EC_KEY_new_method` takes, so a loaded key keeps the engine (and its HSM session) alive for as long as it lives.
- **Applications may make the engine the process-wide default** (the `openssl` apps call `ENGINE_set_default(e, ENGINE_METHOD_ALL)` for `-engine`), which hands the engine's EC method to plain *software* EC keys too. The trampoline therefore asks the handler whether it **owns** the key (an HSM key in ex_data) and delegates unowned keys to OpenSSL's built-in software `sign_sig` — captured when the method is built, precomputed nonces included.

## Provider parity & call paths

Mirrors the provider's raw-ECDSA flow (`azihsm_ossl_signature_ec.c`: pre-hashed input → `AZIHSM_ALGO_ID_ECDSA` → raw `r‖s` → DER), one layer lower — the engine builds the `ECDSA_SIG` and lets OpenSSL DER-encode it.

Because loaded keys adopt the engine's EC method at creation, **both entry points funnel through `sign_sig`**:
- **ABI**: `EVP_DigestSign*` (hashes then signs) and `EVP_PKEY_sign` (pre-hashed).
- **CLI**: `openssl dgst -sign …`, `openssl pkeyutl -sign …`.

OpenSSL's EVP layer hashes *above* the `EC_KEY_METHOD`, so no streaming/`digest_sign` is needed here, and **verify stays in software** using the public key already on the `EVP_PKEY`.

## Testing

Full engine suite on OpenSSL 1.1.x (Ubuntu 20.04 / 1.1.1f): build, clippy `-D warnings` (engine + `azihsm_crypto`), unit tests, doctests, `openssl engine -t`.

- **Unit (mock)** — round-trip `sign_through_loaded_key_verifies`: load an EC key → sign a digest through it via `EVP_DigestSign*` → verify the signature against the public half in software. The load and sign round-trip bodies are factored into a shared `round_trips` module driven against an already-opened `EngineData`, so the mock tests and the hardware smokes run the same flow and differ only in how they open the HSM.
- **Unit (engine-core)** — the sign trampoline is covered directly: null digest / null key / precomputed-nonce rejection, dispatch through a stub handler, and delegation of an unowned software key to OpenSSL's built-in `sign_sig`.
- **Integration (CLI)** — `sign.sh` stages a masked key and signs through the engine via both entry points — `dgst -sign` and `pkeyutl -sign` on the raw digest — verifies both signatures in software with the extracted public key, and checks a tampered message fails verification.
- **Integration (ABI)** — `sign_ec_key_via_engine_capi` loads the key through `ENGINE_load_private_key`, signs via `EVP_DigestSign` **after releasing the test's own ENGINE handle** (locking in the engine-reference lifetime guarantee above), verifies via `EVP_DigestVerify` on the same `EVP_PKEY`, and checks a tampered message fails.

## Testing on hardware

The sign path can be proven on a real device via `sign_ec_key_from_env_smoke` (alongside `open_from_env_smoke` and `load_ec_key_from_env_smoke`). These live in `context/tests.rs` `hw_tests` — compiled only in non-mock builds and `#[ignore]`d, so they never run in the mock CI cell.

On a provisioned HSM host (with OpenSSL 1.1.x on `PKG_CONFIG_PATH`):

```bash
# Credentials matching the partition (32 hex chars = 16 bytes each):
export AZIHSM_CREDENTIALS_ID=<32 hex>  AZIHSM_CREDENTIALS_PIN=<32 hex>

# Resiliency storage dir: must exist, mode 0700, owned by you:
export AZIHSM_RESILIENCY_ENABLED=1
export AZIHSM_RESILIENCY_STORAGE_DIR=/var/lib/azihsm/resiliency
sudo install -d -m 700 -o "$USER" "$AZIHSM_RESILIENCY_STORAGE_DIR"
umask 0077

# OBK/POTA source — TPM (no key files needed):
export AZIHSM_OBK_SOURCE=tpm  AZIHSM_POTA_SOURCE=tpm
# …or caller-supplied keymat:
#   export AZIHSM_OBK_SOURCE=caller  AZIHSM_POTA_SOURCE=caller
#   export AZIHSM_OBK_PATH=$PWD/obk.bin  AZIHSM_MOBK_PATH=$PWD/mobk.bin
#   export AZIHSM_POTA_PRIVATE_KEY_PATH=$PWD/pota_priv.der
#   export AZIHSM_POTA_PUBLIC_KEY_PATH=$PWD/pota_pub.der
#   openssl rand -out obk.bin 48
#   openssl genpkey -algorithm EC -pkeyopt ec_paramgen_curve:secp384r1 -out pota.pem
#   openssl pkcs8 -topk8 -nocrypt -in pota.pem -outform DER -out pota_priv.der
#   openssl pkey -in pota_priv.der -inform DER -pubout -outform DER -out pota_pub.der

# Non-mock build (--features engine, NOT mock), --ignored to run the smokes:
cargo test -p azihsm_ossl_engine --features engine \
  open_from_env_smoke load_ec_key_from_env_smoke sign_ec_key_from_env_smoke \
  -- --ignored --nocapture
```

## Not in scope
ECDH derive (needs an output-shape decision — the SDK returns a key handle/masked blob, not a raw secret) and RSA signing land in follow-ups.


plugins/ossl_engine/azihsm_ossl_engine/src/context.rs
```diff
@@ -525,15 +525,51 @@ fn new_test_engine() -> (Engine, *mut ffi::ENGINE) {
...
-    use std::fs;
-    use std::path::PathBuf;
-    use std::sync::atomic::AtomicU64;
-    use std::sync::atomic::Ordering;
+/// Sign `msg` (SHA-384) through `pkey` via OpenSSL's EVP interface — the same
+/// path the ABI and `openssl dgst -sign` take. Exercises the loaded key's
+/// `EC_KEY_METHOD` `sign_sig` (→ HSM). Shared by the mock and hardware sign tests.
+#[cfg(test)]
+#[allow(unsafe_code)]
+fn evp_digest_sign_sha384(pkey: *mut ffi::EVP_PKEY, msg: &[u8]) -> Vec<u8> {
```
rajesh-gali: This file has grown too big, moving test to seperate file would benefit readability. Also, I see some localized use statements within the functions.

plugins/ossl_engine/azihsm_ossl_engine_core/src/engine.rs
```diff
@@ -263,6 +264,98 @@ unsafe fn load_privkey_inner<H: LoadPrivKeyHandler>(
     H::load(&engine, key_id)
 }
 
+/// Caller-supplied ECDSA signing for an engine-backed EC key, invoked through
+/// the `EC_KEY_METHOD` `sign_sig` hook. Implement on a marker type and pass it
+/// to [`new_ecdsa_method`].
+pub trait EcdsaSignHandler {
```
rajesh-gali: not sure this trait falls into engine.rs which has Engine impl. Should we make this as modular?

plugins/ossl_engine/azihsm_ossl_engine_core/src/engine.rs
```diff
@@ -263,6 +264,98 @@ unsafe fn load_privkey_inner<H: LoadPrivKeyHandler>(
...
+    // SAFETY: method is a fresh EC_KEY_METHOD; get_sign writes the out-params it
+    // is given (NULL for the sign_sig slot, which we don't need).
+    unsafe { ffi::EC_KEY_METHOD_get_sign(method, &mut sign, &mut sign_setup, null_mut()) };
+    // SAFETY: method is ours; reinstall the defaults plus our sign_sig trampoline.
+    unsafe { ffi::EC_KEY_METHOD_set_sign(method, sign, sign_setup, Some(c_ecdsa_sign_sig::<H>)) };
+    Ok(method)
+}
+
 #[cfg(test)]
 mod tests {
```
rajesh-gali, line 281: missing tests for sign round trip and negative tests for ecdsa sign inner, like null eckey, null dgst etc

plugins/ossl_engine/azihsm_ossl_engine_sys/build.rs
```diff
@@ -82,6 +82,11 @@ fn main() {
         .allowlist_function("EC_KEY_.*")
         .allowlist_function("EC_POINT_.*")
         .allowlist_function("EC_GROUP_.*")
+        // ECDSA signature marshaling for the EC_KEY_METHOD sign path: build an
```
rajesh-gali: comments looks like out of place, either we should have single line comment for all entries or none.

plugins/ossl_engine/azihsm_ossl_engine_core/src/engine.rs
```diff
@@ -263,6 +264,98 @@ unsafe fn load_privkey_inner<H: LoadPrivKeyHandler>(
...
+///
+/// # Safety
+/// Called only by OpenSSL's ECDSA sign path. `dgst`/`dgst_len` describe the
+/// digest and `eckey` is the signing key, per the `sign_sig` contract.
+#[allow(unsafe_code)]
+unsafe extern "C" fn c_ecdsa_sign_sig<H: EcdsaSignHandler>(
+    dgst: *const c_uchar,
+    dgst_len: c_int,
+    _in_kinv: *const ffi::BIGNUM,
+    _in_r: *const ffi::BIGNUM,
```
rajesh-gali: I don't think we should ignore caller supplied in_r and kinv silently, api should return unsupported error, otherwise caller will never know these params have been ignored silently

plugins/ossl_engine/azihsm_ossl_engine/src/sign.rs
```diff
@@ -0,0 +1,135 @@
...
+        return Ok(*m as *const ffi::EC_KEY_METHOD);
+    }
+    // Serialize construction so two concurrent first-loads can't each build a
+    // method (only the first is kept; a second would leak).
+    let _guard = INIT.lock();
+    if let Some(m) = METHOD.get() {
+        return Ok(*m as *const ffi::EC_KEY_METHOD);
+    }
+    let method = new_ecdsa_method::<AzihsmEcdsaSign>()?;
+    let _ = METHOD.set(method as usize);
```
rajesh-gali, line 136: what happens when set method returns error? doesn't function returns a method which isn't stored?

walterchris: set cannot fail, so the returned method is always the stored one.

rajesh-gali (approved):

PR #602, opened 2026-07-27
uno: RsaUnwrap bring-up — OAEP endianness fix + non-CRT RSA private-key import

radutta99:
## Summary

Brings up the RSA paths of `RsaUnwrap` on the **uno** HSM firmware. Three changes:

1. **OAEP endianness fix** — `rsa_oaep_decrypt` now flips the `mod_exp_priv` result LE→BE before the RFC 8017 EME-OAEP decode. The uno PKA is little-endian native (LE ciphertext in, LE result out), but the OAEP encoded message `0x00 ‖ maskedSeed ‖ maskedDB` is big-endian; the std PAL does the same flip around OpenSSL. Without it, every RsaUnwrap failed with `RsaDecryptFailed (0x08700012)`. This fixes the AES-key unwrap path.

2. **Non-CRT RSA private-key import** — implements `rsa_priv_der_to_vault`. Adds a small `no_std` DER parser (`der_tlv`, `der_int`, `parse_rsa_priv_der`) handling PKCS#8 `PrivateKeyInfo` and bare PKCS#1 `RSAPrivateKey`. The PKCS#8 `AlgorithmIdentifier` is validated to be `rsaEncryption` (OID 1.2.840.113549.1.1.1 + NULL params), mirroring mcr-hsm's `Asn1RsaEncryptionInfo`. The parsed `n/e/d` are assembled into the PKA vault operand `[d(k) ‖ n(k) ‖ e(4)]` little-endian — matching mcr-hsm's `RsaPrivKey::to_pka_bytes` and what `rsa_priv_pub_key` reads back. The stack scratch holding the private exponent is scrubbed with `zeroize`. CRT import returns `UnsupportedCmd` (deferred — needs PKA-derived `n1q`/`n2p`).

3. **OAEP oversized-output error contract** — `rsa_oaep_decrypt` now returns `RsaInvalidKeyLength` (not `InvalidArg`) when the recovered message exceeds the caller's output buffer, matching the std PAL. The shared key-unwrap path maps this to `RsaUnwrapInvalidKek`, so an oversized recovered KEK is now reported correctly.

## Testing

**Hardware (Manticore EVB):**
- `rsa_unwrap_generated_key`: 12/17 (was 9/17)
- `rsa_unwrap_smoke`: 3/5 (was 2/5) — `oversized_kek_smoke` now passes
- `get_unwrapping_key`: 7/7 (unchanged)

**Emu:** `precheck --nextest --package azihsm_ddi_mbor_types --features emu --filter smoke --profile ci-emu-smoke` → 94/94 passed.

fmt / clippy (`-D warnings`, ms-nightly) / taplo all clean.

## Deferred to follow-up PRs

The remaining `rsa_unwrap` failures each map cleanly to one of:
- **CRT RSA import** — needs PKA-derived `n1q`/`n2p` (mcr-hsm computes these separately).
- **uno ECC key import** — blocks `ecc_key_smoke`, `ecc_keys_with_key_tag`.
- **OpenKey-by-tag** — a full SDK-wide DDI (op 1015 has no handler); needs a shared PAL-trait + vault change (the uno vault `Entry` already carries a `session_or_tag` field).

## Base

Stacked on #601 (`user/radutta/get-unwrapping-key`), which is not yet merged — this PR targets that branch, not `main`.

### DIFF
--- a/fw/plat/uno/fw/pal/Cargo.toml
+++ b/fw/plat/uno/fw/pal/Cargo.toml
@@ -46,7 +46,10 @@ embassy-time = { workspace = true }
 tock-registers = { workspace = true }
 
 bitfield-struct = { workspace = true }
-der = { version = "0.8", default-features = false, features = ["derive", "oid"] }
+der = { default-features = false, features = [
+  "derive",
+  "oid",
+], version = "0.8" }
 open-enum = { workspace = true }
 zerocopy = { workspace = true }
 zeroize = { workspace = true }
--- a/fw/plat/uno/fw/pal/Cargo.toml
+++ b/fw/plat/uno/fw/pal/Cargo.toml
@@ -46,10 +46,7 @@ embassy-time = { workspace = true }
 tock-registers = { workspace = true }
 
 bitfield-struct = { workspace = true }
-der = { default-features = false, features = [
-  "derive",
-  "oid",
-], version = "0.8" }
+der = { workspace = true }
 open-enum = { workspace = true }
 zerocopy = { workspace = true }
 zeroize = { workspace = true }
--- a/fw/plat/uno/fw/pal/Cargo.toml
+++ b/fw/plat/uno/fw/pal/Cargo.toml
@@ -46,6 +46,7 @@ embassy-time = { workspace = true }
 tock-registers = { workspace = true }
 
 bitfield-struct = { workspace = true }
+der = { version = "0.8", default-features = false, features = ["derive", "oid"] }
 open-enum = { workspace = true }
 zerocopy = { workspace = true }
 zeroize = { workspace = true }
--- a/fw/plat/uno/fw/pal/Cargo.toml
+++ b/fw/plat/uno/fw/pal/Cargo.toml
@@ -48,6 +48,7 @@ tock-registers = { workspace = true }
 bitfield-struct = { workspace = true }
 open-enum = { workspace = true }
 zerocopy = { workspace = true }
+zeroize = { workspace = true }
 
 [features]
 default = []
--- a/fw/plat/uno/fw/pal/src/asn1.rs
+++ b/fw/plat/uno/fw/pal/src/asn1.rs
@@ -73,7 +73,7 @@ pub(crate) struct RsaPrivateKeyAsn1<'a> {
     pub(crate) coefficient: UintRef<'a>,
 }
 
-/// Decodes a recovered RSA private key — a PKCS#8 `PrivateKeyInfo` or a bare
+/// Decodes a recovered RSA private key — a PKCS#8 `PrivateKeyInfo` wrapping a
 /// PKCS#1 `RSAPrivateKey` — into the validated [`RsaPrivateKeyAsn1`].
 ///
 /// The `der` crate enforces canonical DER (definite minimal-length encodings,
@@ -83,17 +83,17 @@ pub(crate) struct RsaPrivateKeyAsn1<'a> {
 /// sizes (modulus width, exponent width) are a PKA concern and are checked by
 /// the caller during operand assembly.
 pub(crate) fn parse_rsa_private_key(der_bytes: &[u8]) -> Option<RsaPrivateKeyAsn1<'_>> {
-    let key = if let Ok(pki) = RsaPrivateKeyInfo::from_der(der_bytes) {
-        // RFC 5208 §5: PrivateKeyInfo `version` is v1 (= 0). RFC 5958 adds v2
-        // (= 1) with public-key / attributes, which this parser does not model
-        // — reject rather than silently drop the extra fields.
-        if pki.version != 0 || pki.algorithm.oid != RSA_ENCRYPTION {
-            return None;
-        }
-        RsaPrivateKeyAsn1::from_der(pki.private_key.as_bytes()).ok()?
-    } else {
-        RsaPrivateKeyAsn1::from_der(der_bytes).ok()?
-    };
+    // The recovered wire key is always a PKCS#8 PrivateKeyInfo (the format the
+    // RsaUnwrap collateral is generated in); a bare PKCS#1 RSAPrivateKey is not
+    // accepted.
+    let pki = RsaPrivateKeyInfo::from_der(der_bytes).ok()?;
+    // RFC 5208 §5: PrivateKeyInfo `version` is v1 (= 0). RFC 5958 adds v2
+    // (= 1) with public-key / attributes, which this parser does not model
+    // — reject rather than silently drop the extra fields.
+    if pki.version != 0 || pki.algorithm.oid != RSA_ENCRYPTION {
+        return None;
+    }
+    let key = RsaPrivateKeyAsn1::from_der(pki.private_key.as_bytes()).ok()?;
     // RFC 8017 §A.1.2: `version` is 0 for two-prime keys. Version 1 (multi-prime)
     // would require `otherPrimeInfos`, which our SEQUENCE does not declare.
     if key.version != 0 {
--- a/fw/plat/uno/fw/pal/src/asn1.rs
+++ b/fw/plat/uno/fw/pal/src/asn1.rs
@@ -1,0 +1,103 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! ASN.1 / DER decoders for imported private keys, built on the no-alloc
+//! [`der`] crate (the same dependency `fw/core/crypto/x509-chain` already uses).
+//!
+//! The wrapped-key import paths (`RsaUnwrap`) recover a plaintext DER private
+//! key and must decode it into the raw components the PKA vault operand is
+//! assembled from. This module owns only the **pure ASN.1 decode** (PKCS#8 / PKCS#1, and — as they land — SEC1 EC); the
+//! platform-specific little-endian PKA operand layout lives in the
+//! per-algorithm modules ([`rsa`](crate::crypto::rsa), [`ecc`](crate::crypto::ecc)) so this
+//! stays a small, hardware-agnostic parsing layer.
+//!
+//! ## Scope and forward-looking extension points
+//!
+//! - **RSA non-CRT (today):** [`parse_rsa_private_key`] →
+//!   [`RsaPrivateKeyAsn1`]; `rsa` assembles `[d ‖ n ‖ e]`.
+//! - **RSA CRT (#608):** reuses [`RsaPrivateKeyAsn1`] as-is — it already decodes
+//!   the CRT fields (`prime1`, `prime2`, `exponent1`, `exponent2`,
+//!   `coefficient`). Only the CRT operand assembly (and the derived `n1q`/`n2p`
+//!   PKA math) is added in `rsa`; **no change is needed here**.
+//! - **ECC (#604):** add the SEC1 `ECPrivateKey` / PKCS#8 EC `PrivateKeyInfo`
+//!   decoders alongside the RSA ones below, plus a `parse_ec_private_key`
+//!   returning the scalar + curve OID; consumed by `ecc`.
+
+use der::Decode;
+use der::Sequence;
+use der::asn1::Null;
+use der::asn1::ObjectIdentifier;
+use der::asn1::OctetStringRef;
+use der::asn1::UintRef;
+
+// ── RSA (PKCS#8 PrivateKeyInfo / PKCS#1 RSAPrivateKey) ─────────────────────
+
+/// rsaEncryption OID (1.2.840.113549.1.1.1) — the algorithm identifier of an
+/// RSA key inside a PKCS#8 `PrivateKeyInfo`.
+const RSA_ENCRYPTION: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.1");
+
+/// PKCS#8 `AlgorithmIdentifier` for RSA: the `rsaEncryption` OID with the
+/// explicit `NULL` parameters required by RFC 3279 §2.3.1.
+#[derive(Sequence)]
+struct RsaAlgorithmIdentifier {
+    oid: ObjectIdentifier,
+    parameters: Null,
+}
+
+/// PKCS#8 `PrivateKeyInfo` (RFC 5208) whose `privateKey` OCTET STRING wraps a
+/// PKCS#1 `RSAPrivateKey`.
+#[derive(Sequence)]
+struct RsaPrivateKeyInfo<'a> {
+    version: u8,
+    algorithm: RsaAlgorithmIdentifier,
+    private_key: &'a OctetStringRef,
+}
+
+/// PKCS#1 `RSAPrivateKey` (RFC 8017 A.1.2), two-prime form (version 0).
+///
+/// The full SEQUENCE is decoded so the `der` reader validates every field. The
+/// non-CRT import path reads only `modulus` / `public_exponent` /
+/// `private_exponent`; the CRT import path (#608) additionally reads the CRT
+/// quintuple (`prime1`, `prime2`, `exponent1`, `exponent2`, `coefficient`),
+/// which is why they are decoded and exposed here rather than skipped.
+#[derive(Sequence)]
+pub(crate) struct RsaPrivateKeyAsn1<'a> {
+    pub(crate) version: u8,
+    pub(crate) modulus: UintRef<'a>,
+    pub(crate) public_exponent: UintRef<'a>,
+    pub(crate) private_exponent: UintRef<'a>,
+    pub(crate) prime1: UintRef<'a>,
+    pub(crate) prime2: UintRef<'a>,
+    pub(crate) exponent1: UintRef<'a>,
+    pub(crate) exponent2: UintRef<'a>,
+    pub(crate) coefficient: UintRef<'a>,
+}
+
+/// Decodes a recovered RSA private key — a PKCS#8 `PrivateKeyInfo` or a bare
+/// PKCS#1 `RSAPrivateKey` — into the validated [`RsaPrivateKeyAsn1`].
+///
+/// The `der` crate enforces canonical DER (definite minimal-length encodings,
+/// no trailing bytes via `from_der`) and rejects negative / non-minimal
+/// INTEGERs. On top of that this rejects key versions the two-prime SEQUENCE
+/// does not model, so callers get a fully-structurally-validated key. Field
+/// sizes (modulus width, exponent width) are a PKA concern and are checked by
+/// the caller during operand assembly.
+pub(crate) fn parse_rsa_private_key(der_bytes: &[u8]) -> Option<RsaPrivateKeyAsn1<'_>> {
+    let key = if let Ok(pki) = RsaPrivateKeyInfo::from_der(der_bytes) {
+        // RFC 5208 §5: PrivateKeyInfo `version` is v1 (= 0). RFC 5958 adds v2
+        // (= 1) with public-key / attributes, which this parser does not model
+        // — reject rather than silently drop the extra fields.
+        if pki.version != 0 || pki.algorithm.oid != RSA_ENCRYPTION {
+            return None;
+        }
+        RsaPrivateKeyAsn1::from_der(pki.private_key.as_bytes()).ok()?
+    } else {
+        RsaPrivateKeyAsn1::from_der(der_bytes).ok()?
+    };
+    // RFC 8017 §A.1.2: `version` is 0 for two-prime keys. Version 1 (multi-prime)
+    // would require `otherPrimeInfos`, which our SEQUENCE does not declare.
+    if key.version != 0 {
+        return None;
+    }
+    Some(key)
+}
--- a/fw/plat/uno/fw/pal/src/crypto/rsa.rs
+++ b/fw/plat/uno/fw/pal/src/crypto/rsa.rs
@@ -441,6 +441,13 @@ impl HsmRsa for UnoHsmPal {
         self.mod_exp_priv(io, key_size, priv_key, ciphertext, em)
             .await?;
 
+        // `mod_exp_priv` is little-endian (PKA-native): both the ciphertext
+        // operand and the `em` result are LE wire form. The RSA-OAEP decode
+        // below (RFC 8017 EME-OAEP) operates on the big-endian encoded message
+        // `EM = 0x00 ‖ maskedSeed ‖ maskedDB`, so flip the result LE→BE first
+        // (the std PAL does the same flip around OpenSSL, in its driver).
+        em[..k].reverse();
+
         if em[0] != 0x00 {
             return Err(HsmError::RsaDecryptFailed);
         }
--- a/fw/plat/uno/fw/pal/src/crypto/rsa.rs
+++ b/fw/plat/uno/fw/pal/src/crypto/rsa.rs
@@ -23,15 +23,11 @@ use azihsm_fw_hsm_pal_traits::HsmRsaKey;
 use azihsm_fw_hsm_pal_traits::HsmRsaPct;
 use azihsm_fw_hsm_pal_traits::HsmScopedAlloc;
 use azihsm_fw_uno_drivers_upka::UpkaRsaKeyType;
-use der::Decode;
-use der::Sequence;
-use der::asn1::Null;
-use der::asn1::ObjectIdentifier;
-use der::asn1::OctetStringRef;
-use der::asn1::UintRef;
 use zeroize::Zeroize;
 
 use crate::UnoHsmPal;
+use crate::asn1::RsaPrivateKeyAsn1;
+use crate::asn1::parse_rsa_private_key;
 
 // =============================================================================
 // Helper functions
@@ -75,78 +71,18 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
 }
 
 // =============================================================================
-// DER parsing for RSA private-key import (PKCS#8 / PKCS#1 RSAPrivateKey)
+// Vault operand assembly (imported RSA private key → PKA layout)
 // =============================================================================
 
 /// Maximum RSA modulus length in bytes (RSA-4096).
 const MAX_RSA_MODULUS_LEN: usize = 512;
 
-/// rsaEncryption OID (1.2.840.113549.1.1.1) — the algorithm identifier of an
-/// RSA key inside a PKCS#8 `PrivateKeyInfo`.
-const RSA_ENCRYPTION: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.1");
-
-/// PKCS#8 `AlgorithmIdentifier` for RSA: the `rsaEncryption` OID with the
-/// explicit `NULL` parameters required by RFC 3279 §2.3.1.
-#[derive(Sequence)]
-struct RsaAlgorithmIdentifier {
-    algorithm: ObjectIdentifier,
-    parameters: Null,
-}
-
-/// PKCS#8 `PrivateKeyInfo` (RFC 5208) whose `privateKey` OCTET STRING wraps a
-/// PKCS#1 `RSAPrivateKey`.
-#[derive(Sequence)]
-struct RsaPrivateKeyInfo<'a> {
-    version: u8,
-    algorithm: RsaAlgorithmIdentifier,
-    private_key: &'a OctetStringRef,
-}
-
-/// PKCS#1 `RSAPrivateKey` (RFC 8017 A.1.2), two-prime form (version 0). The full
-/// SEQUENCE is decoded so the `der` reader validates every field; only
-/// `modulus`, `public_exponent`, and `private_exponent` feed the non-CRT vault
-/// operand.
-#[derive(Sequence)]
-struct RsaPrivateKeyDer<'a> {
-    version: u8,
-    modulus: UintRef<'a>,
-    public_exponent: UintRef<'a>,
-    private_exponent: UintRef<'a>,
-    prime1: UintRef<'a>,
-    prime2: UintRef<'a>,
-    exponent1: UintRef<'a>,
-    exponent2: UintRef<'a>,
-    coefficient: UintRef<'a>,
-}
-
-/// Decodes a recovered RSA private key — a PKCS#8 `PrivateKeyInfo` or a bare
-/// PKCS#1 `RSAPrivateKey` — and assembles the little-endian PKA vault operand
-/// `[d(k) ‖ n(k) ‖ e(4)]` into `out`, returning `(vault_len, modulus_len)`.
-///
-/// The `der` crate enforces canonical DER (definite minimal-length encodings, no
-/// trailing bytes via `from_der`) and rejects negative / non-minimal INTEGERs,
-/// so this carries no hand-rolled TLV / INTEGER validation. Reads `der_bytes`
-/// only; the caller scrubs the recovered plaintext on any error.
-fn rsa_der_to_vault_operand(der_bytes: &[u8], out: &mut [u8]) -> Option<(usize, usize)> {
-    // Assemble inside each arm while the decoded `RSAPrivateKey` — and the
-    // `UintRef`s borrowing `der_bytes` — is still alive.
-    if let Ok(pki) = RsaPrivateKeyInfo::from_der(der_bytes) {
-        if pki.algorithm.algorithm != RSA_ENCRYPTION {
-            return None;
-        }
-        let inner = RsaPrivateKeyDer::from_der(pki.private_key.as_bytes()).ok()?;
-        assemble_rsa_operand(&inner, out)
-    } else {
-        let inner = RsaPrivateKeyDer::from_der(der_bytes).ok()?;
-        assemble_rsa_operand(&inner, out)
-    }
-}
-
 /// Validates the RSA field sizes and writes `[d ‖ n ‖ e]` little-endian into
 /// `out`, returning `(vault_len, modulus_len)`. Rejects moduli that are not
 /// 2048/3072/4096-bit, an exponent wider than 4 bytes, or a `d` wider than the
-/// modulus.
-fn assemble_rsa_operand(key: &RsaPrivateKeyDer<'_>, out: &mut [u8]) -> Option<(usize, usize)> {
+/// modulus. Structural validation (versions, algorithm OID) is done upstream by
+/// [`parse_rsa_private_key`](super::asn1::parse_rsa_private_key).
+fn assemble_rsa_operand(key: &RsaPrivateKeyAsn1<'_>, out: &mut [u8]) -> Option<(usize, usize)> {
     let n = key.modulus.as_bytes();
     let e = key.public_exponent.as_bytes();
     let d = key.private_exponent.as_bytes();
@@ -279,10 +215,13 @@ impl HsmRsa for UnoHsmPal {
             return Err(HsmError::UnsupportedCmd);
         }
         // Decode + assemble the little-endian vault operand `[d(k) ‖ n(k) ‖ e(4)]`
-        // into a scratch buffer. This only reads `buf` (the recovered plaintext
-        // DER); the borrow ends before the in-place rewrite below.
+        // into a scratch buffer. Only `buf` is read (the recovered plaintext
+        // DER); the decoded `UintRef`s alias into it, and the borrow ends
+        // before the in-place rewrite below.
         let mut out = [0u8; MAX_RSA_MODULUS_LEN * 2 + 4];
-        let Some((vault_len, modulus_len)) = rsa_der_to_vault_operand(&buf[..], &mut out) else {
+        let assembled =
+            parse_rsa_private_key(&buf[..]).and_then(|key| assemble_rsa_operand(&key, &mut out));
+        let Some((vault_len, modulus_len)) = assembled else {
             buf.zeroize();
             out.zeroize();
             return Err(HsmError::InvalidArg);
--- a/fw/plat/uno/fw/pal/src/crypto/rsa.rs
+++ b/fw/plat/uno/fw/pal/src/crypto/rsa.rs
@@ -23,6 +23,12 @@ use azihsm_fw_hsm_pal_traits::HsmRsaKey;
 use azihsm_fw_hsm_pal_traits::HsmRsaPct;
 use azihsm_fw_hsm_pal_traits::HsmScopedAlloc;
 use azihsm_fw_uno_drivers_upka::UpkaRsaKeyType;
+use der::Decode;
+use der::Sequence;
+use der::asn1::Null;
+use der::asn1::ObjectIdentifier;
+use der::asn1::OctetStringRef;
+use der::asn1::UintRef;
 use zeroize::Zeroize;
 
 use crate::UnoHsmPal;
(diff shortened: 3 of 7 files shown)


fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -67,6 +68,118 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
...
+    let (tag, mut start, mut len, next) = der_tlv(der, pos)?;
+    if tag != 0x02 || len == 0 {
+        return None;
+    }
+    while len > 1 && der[start] == 0x00 {
+        start += 1;
+        len -= 1;
+    }
+    Some((start, len, next))
+}
```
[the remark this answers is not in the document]
radutta99: Good catch. Hardened der_int in 694d7614: it now rejects negative DER INTEGERs (MSB set on the first content byte with no

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -441,6 +577,13 @@ impl HsmRsa for UnoHsmPal {
         self.mod_exp_priv(io, key_size, priv_key, ciphertext, em)
             .await?;
 
+        // `mod_exp_priv` is little-endian (PKA-native): both the ciphertext
+        // operand and the `em` result are LE wire form. The RSA-OAEP decode
+        // below (RFC 8017 EME-OAEP) operates on the big-endian encoded message
+        // `EM = 0x00 ‖ maskedSeed ‖ maskedDB`, so flip the result LE→BE first
+        // (the std PAL does the same flip around OpenSSL, in its driver).
```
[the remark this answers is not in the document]
radutta99: Fixed in 694d7614. 
sa_oaep_decrypt now scrubs the secret-bearing m scratch with DmaBuf::zeroize on every exit path: the fallible OAEP decode (mod_exp_priv, MGF recovery, lHash check, separator scan, output copy) runs in an inner block, and a single m.zeroize() runs after it on both the success and all ?-propagated / early-return error paths before the scoped allocator rewinds. HW-verified: no regression (15/22).

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -160,14 +288,37 @@ impl HsmRsa for UnoHsmPal {
...
+        let vault_len = 2 * modulus_len + 4;
+        let mut out = [0u8; MAX_RSA_MODULUS_LEN * 2 + 4];
+        write_le(&mut out[0..modulus_len], &buf[ds..ds + dl]);
+        write_le(&mut out[modulus_len..2 * modulus_len], &buf[ns..ns + nl]);
+        write_le(&mut out[2 * modulus_len..vault_len], &buf[es..es + el]);
+        buf[..vault_len].copy_from_slice(&out[..vault_len]);
+        // Scrub the assembled operand (holds the private exponent `d`) from the
+        // stack scratch. `Zeroize` uses volatile writes so the wipe is not
+        // elided as a dead store.
+        out.zeroize();
```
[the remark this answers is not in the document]
radutta99: Fixed in 7d3a27db. rsa_priv_der_to_vault now zeroizes buf[vault_len..] after rewriting the vault operand, so the leftover source DER (which for a full RSAPrivateKey still holds the secret CRT components p/q/dp/dq/qinv) is scrubbed before the scoped DMA buffer is reused. HW-verified: rsa_key + rsa_key_sizes still pass (12/17, unchanged).

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -67,6 +68,133 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
...
+
+/// Parses a DER RSA private key — PKCS#8 `PrivateKeyInfo` or a bare PKCS#1
+/// `RSAPrivateKey` — and returns the big-endian value ranges `(start, len)` of
+/// the modulus `n`, public exponent `e`, and private exponent `d`.
+fn parse_rsa_priv_der(der: &[u8]) -> Option<RsaDerRanges> {
+    // Outer SEQUENCE.
+    let (tag, seq_start, _seq_len, _next) = der_tlv(der, 0)?;
+    if tag != 0x30 {
+        return None;
+    }
```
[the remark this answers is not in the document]
radutta99: Fixed in 7d3a27db. parse_rsa_priv_der now rejects trailing garbage: the outer SEQUENCE must span the entire input (outer_next == der.len()), and the PKCS#8 privateKey OCTET STRING must wrap exactly the inner RSAPrivateKey SEQUENCE (inner_next == oct_next). This is safe because the recovered material is exact-length (unwrap_key returns payload_buf[..payload_len]). HW-verified: rsa_key + rsa_key_sizes + incorrect_der_len still pass.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -67,6 +68,135 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
...
+        let n = (len_byte & 0x7f) as usize;
+        if n == 0 || n > 4 {
+            return None;
+        }
+        let mut len = 0usize;
+        for i in 0..n {
+            len = (len << 8) | (*der.get(pos + 2 + i)? as usize);
+        }
+        (len, 2 + n)
+    };
```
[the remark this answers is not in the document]
radutta99: Fixed in 00e9b177. der_tlv now rejects non-minimal long-form definite lengths: the most-significant length byte must be non-zero (no leading-zero padding), and long form must not encode a length that fits in short form (len must be >= 0x80). This matches strict DER decoders for the untrusted recovered key material. HW-verified: rsa_key + rsa_key_sizes still pass (12/17, unchanged).

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -160,14 +299,42 @@ impl HsmRsa for UnoHsmPal {
...
+        }
+        // Parse the recovered DER (PKCS#8 or bare PKCS#1) into the big-endian
+        // value ranges of the modulus `n`, public exponent `e`, and private
+        // exponent `d`. The ranges alias `buf`.
+        let ((ns, nl), (es, el), (ds, dl)) =
+            parse_rsa_priv_der(&buf[..]).ok_or(HsmError::InvalidArg)?;
+        let modulus_len = nl;
+        if !matches!(modulus_len, 256 | 384 | 512) || el > 4 || dl > modulus_len {
+            return Err(HsmError::InvalidArg);
+        }
```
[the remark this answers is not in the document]
radutta99: Fixed in e1cd9fa2. rsa_priv_der_to_vault now zeroizes buf on every early-error return (crt==true, DER parse failure, and size validation), so the recovered plaintext RSA private-key DER is scrubbed from the reused DMA SRAM after a failed import. The success path is unchanged: it overwrites buf[..vault_len] with the vault operand and scrubs the buf[vault_len..] tail.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -160,14 +299,49 @@ impl HsmRsa for UnoHsmPal {
...
+        // Assemble the vault operand `[d(k) ‖ n(k) ‖ e(4)]` little-endian in a
+        // scratch buffer, then write it back in place — the vault form is
+        // smaller than the source DER, so the overwrite is safe.
+        let vault_len = 2 * modulus_len + 4;
+        let mut out = [0u8; MAX_RSA_MODULUS_LEN * 2 + 4];
+        write_le(&mut out[0..modulus_len], &buf[ds..ds + dl]);
+        write_le(&mut out[modulus_len..2 * modulus_len], &buf[ns..ns + nl]);
+        write_le(&mut out[2 * modulus_len..vault_len], &buf[es..es + el]);
+        buf[..vault_len].copy_from_slice(&out[..vault_len]);
```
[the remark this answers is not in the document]
radutta99: Fixed in 0f88cf98. rsa_priv_der_to_vault now checks vault_len <= buf.len() before the in-place slices and, on failure, scrubs buf and returns InvalidArg. This prevents an out-of-bounds panic on malformed/truncated PKCS#1 (e.g. a very short d) where 2*modulus_len+4 could exceed the recovered buffer length. HW-verified: rsa_key + rsa_key_sizes still pass (12/17).

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -67,6 +68,144 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
...
+        let (_at, alg_start, _acl, after_alg) = der_tlv(der, p)?;
+        let (oid_tag, oid_start, oid_len, after_oid) = der_tlv(der, alg_start)?;
+        if oid_tag != 0x06 || der[oid_start..oid_start + oid_len] != RSA_ENCRYPTION_OID {
+            return None;
+        }
+        // The rsaEncryption parameters must be NULL (tag 0x05, length 0).
+        let (null_tag, _null_start, null_len, _null_next) = der_tlv(der, after_oid)?;
+        if null_tag != 0x05 || null_len != 0 {
+            return None;
+        }
```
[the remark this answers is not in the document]
radutta99: Fixed in 0f88cf98. parse_rsa_priv_der now requires the NULL parameters TLV to consume the remainder of the AlgorithmIdentifier SEQUENCE (null_next == after_alg), rejecting any extra trailing fields in the algorithm id. HW-verified: valid PKCS#8 keys (rsa_key, rsa_key_sizes) still import.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -67,6 +68,145 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
...
+        (len_byte as usize, 2)
+    } else {
+        let n = (len_byte & 0x7f) as usize;
+        if n == 0 || n > 4 {
+            return None;
+        }
+        // DER requires minimal length encoding: the most-significant length
+        // byte must be non-zero (no leading-zero padding).
+        if *der.get(pos + 2)? == 0 {
+            return None;
```
vsonims: should we use asn1 crate instead of manually parsing DER?

vsonims (changes requested):

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: used der::asn1 that is being used for x509-chain

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -67,6 +70,45 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
...
+/// modulus. Structural validation (versions, algorithm OID) is done upstream by
+/// [`parse_rsa_private_key`](super::asn1::parse_rsa_private_key).
+fn assemble_rsa_operand(key: &RsaPrivateKeyAsn1<'_>, out: &mut [u8]) -> Option<(usize, usize)> {
+    let n = key.modulus.as_bytes();
+    let e = key.public_exponent.as_bytes();
+    let d = key.private_exponent.as_bytes();
+    let modulus_len = n.len();
+    if !matches!(modulus_len, 256 | 384 | 512) || e.len() > 4 || d.len() > modulus_len {
+        return None;
+    }
```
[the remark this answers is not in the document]
radutta99: `UintRef` is der's *unsigned* INTEGER type. Its `new()` calls `strip_leading_zeroes(...)` and `as_bytes()` is documented as "the least significant bytes of a big endian integer value with all leading zeros stripped" (der 0.8.1 `asn1/integer/uint.rs`). So the DER sign-padding `0x00` on `n`/`d` is removed — `modulus_len` is exactly 256/384/512 for RSA-2048/3072/4096, not 257/513. Its `decode_value` also rejects non-canonical encodings.

HW-verified: `test_rsa_unwrap_rsa_key` and `test_rsa_unwrap_rsa_key_sizes` (2048/3072/4096) pass on the Manticore EVB, so valid keys are accepted. No change needed.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Done. Switched to the RustCrypto `der` crate (the same no-alloc dependency `fw/core/crypto/x509-chain` already uses) with declarative `#[derive(Sequence)]` decoders, and extracted them into a dedicated `pal/src/asn1.rs` module. Commits `76737f4a` (der migration) and `6fa6f59e` (module extraction).

The `der` canonical reader now handles the minimal-length / trailing-byte / non-minimal-INTEGER validation the hand-rolled parser was doing by hand. Note: mcr-hsm's `hsm/src/der` module (built on rust-asn1) uses `alloc::Vec` and the uno PAL is no-alloc, so it can't be reused directly; `der` gives the same declarative approach no-alloc. HW: rsa_unwrap 15/22, unchanged.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -67,6 +70,45 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
...
+/// `out`, returning `(vault_len, modulus_len)`. Rejects moduli that are not
+/// 2048/3072/4096-bit, an exponent wider than 4 bytes, or a `d` wider than the
+/// modulus. Structural validation (versions, algorithm OID) is done upstream by
+/// [`parse_rsa_private_key`](super::asn1::parse_rsa_private_key).
+fn assemble_rsa_operand(key: &RsaPrivateKeyAsn1<'_>, out: &mut [u8]) -> Option<(usize, usize)> {
+    let n = key.modulus.as_bytes();
+    let e = key.public_exponent.as_bytes();
+    let d = key.private_exponent.as_bytes();
+    let modulus_len = n.len();
+    if !matches!(modulus_len, 256 | 384 | 512) || e.len() > 4 || d.len() > modulus_len {
```
jaygmsft: should e len be always 4 and d len be always same as n len? unless these are directly from der currently where the leading zeros get stripped.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -160,14 +202,50 @@ impl HsmRsa for UnoHsmPal {
...
-        // (RsaUnwrap RSA import).
-        Err(HsmError::UnsupportedCmd)
+        // CRT import needs the derived PKA operands (n1q, n2p) computed from
+        // p/q/n via PKA modular arithmetic — not yet brought up on Uno.
+        if crt {
+            // `buf` still holds the recovered plaintext DER; scrub it before
+            // any early return so no secret key material lingers in the reused
+            // DMA SRAM after a failed import.
+            buf.zeroize();
+            return Err(HsmError::UnsupportedCmd);
```
jaygmsft, line 272: Missing a TODO comment?

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -160,14 +202,50 @@ impl HsmRsa for UnoHsmPal {
...
+        }
+        // Decode + assemble the little-endian vault operand `[d(k) ‖ n(k) ‖ e(4)]`
+        // into a scratch buffer. Only `buf` is read (the recovered plaintext
+        // DER); the decoded `UintRef`s alias into it, and the borrow ends
+        // before the in-place rewrite below.
+        let mut out = [0u8; MAX_RSA_MODULUS_LEN * 2 + 4];
+        let assembled =
+            parse_rsa_private_key(&buf[..]).and_then(|key| assemble_rsa_operand(&key, &mut out));
+        let Some((vault_len, modulus_len)) = assembled else {
+            buf.zeroize();
```
jaygmsft: buf should be zeroized when IO finishes, do you need to zeroize still?

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -160,14 +202,50 @@ impl HsmRsa for UnoHsmPal {
...
+            // any early return so no secret key material lingers in the reused
+            // DMA SRAM after a failed import.
+            buf.zeroize();
+            return Err(HsmError::UnsupportedCmd);
+        }
+        // Decode + assemble the little-endian vault operand `[d(k) ‖ n(k) ‖ e(4)]`
+        // into a scratch buffer. Only `buf` is read (the recovered plaintext
+        // DER); the decoded `UintRef`s alias into it, and the borrow ends
+        // before the in-place rewrite below.
+        let mut out = [0u8; MAX_RSA_MODULUS_LEN * 2 + 4];
```
jaygmsft: I think we should avoid allocating such a huge buffer on stack memory. This is >1KB. I see no reason why we cannot do it in buf in place.

fw/plat/uno/fw/pal/src/asn1.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub(crate) private_exponent: UintRef<'a>,
+    pub(crate) prime1: UintRef<'a>,
+    pub(crate) prime2: UintRef<'a>,
+    pub(crate) exponent1: UintRef<'a>,
+    pub(crate) exponent2: UintRef<'a>,
+    pub(crate) coefficient: UintRef<'a>,
+}
+
+/// Decodes a recovered RSA private key — a PKCS#8 `PrivateKeyInfo` or a bare
+/// PKCS#1 `RSAPrivateKey` — into the validated [`RsaPrivateKeyAsn1`].
```
jaygmsft, line 77: Is there any test that needs PKCS#1? If not why do we need it?

fw/plat/uno/fw/pal/Cargo.toml
```diff
@@ -46,8 +46,13 @@ embassy-time = { workspace = true }
 tock-registers = { workspace = true }
 
 bitfield-struct = { workspace = true }
+der = { default-features = false, features = [
```
jaygmsft: External crate added directly. der crate has too many dependencies. Plus has dependency on rust crypto. No go.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Yes, this is coming in #608? No point adding it now since the PR is out already.

fw/plat/uno/fw/pal/Cargo.toml
radutta99: Will bring in ASN.1 create.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Right — these come straight from the `der` `UintRef`s, whose `as_bytes()` strips leading zeros. So `e` is usually 3 bytes (`0x01 00 01`), not always 4, and `d` can be shorter than `n` when its top byte is zero. That's why the checks are upper bounds (`e.len() > 4`, `d.len() > modulus_len`) rather than equality — we accept the canonical minimal encodings and left-zero-pad them into the LE operand via `write_le`.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: No, it does not.  unwrap_key returns a plain slice (&mut palyload_buf[..]) its a view into the bump arena, so no drop on zeroize.  So we need to scrub it. 
(I think it was one of the copilot comments earlier)

jaygmsft: All DMA buf allocations happen from the 8KB buffer passed to the request. That should be zeroized when IO finishes. It definitely happens in std. Please check again. If it doesnt happen for uno then that should be fixed. Also that zeroization should be done with DMA engine so we arent zeroizing the 8KB on cpu.

fw/plat/uno/fw/pal/Cargo.toml
radutta99: ASN.1 crate depends on global allocator unconditionally which uno does not have.

fw/plat/uno/fw/pal/src/asn1.rs
radutta99: Fair point. Removed PKCS#1 references and logic.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Great point! Added fully in-place assembly — no heap, no stack array

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Anyway, added a TODO comment.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: yes, std zeroizes on buf_pool::free. Uno never zeroizes across IO lifecycle. It only rewinds the bump watermark. The only place uno zeroizes implicitly is on allocation using dma_alloc_zeroed.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -67,6 +68,88 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
...
+    }
+    let mut e_arr = [0u8; 4];
+    e_arr[..e.len()].copy_from_slice(e);
+    Some(RsaOperandLayout {
+        modulus_len,
+        n_off: offset_in(buf, n),
+        d_off: offset_in(buf, d),
+        d_len: d.len(),
+        e: e_arr,
+        e_len: e.len(),
```
jaygmsft: if we are passing e_len, why do we need to make e as 4 byte length, and if the length is hardcoded to 4 bytes then why do we need to send e_len? should only need to do one thing.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -438,50 +549,73 @@ impl HsmRsa for UnoHsmPal {
...
+            }
+
+            output[..m_len].copy_from_slice(&db[m_start..]);
+            Ok(m_len)
         }
+        .await;
 
-        output[..m_len].copy_from_slice(&db[m_start..]);
-        Ok(m_len)
+        em.zeroize();
```
jaygmsft: if central zeroize is coming later then we should not have to add zeroize over here. lets not add code that we know we need to remove when it can be avoided.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Will bring it as a separate PR. When the zeroization PR comes is, all the small explicit zeroization here in this PR and elsewhere in other PAL trait adapter functions today will go away

fw/plat/uno/fw/pal/Cargo.toml
radutta99: We decided offline, that DER crate is okay since it doesnt use the default features and bring in only a couple of things.

fw/plat/uno/fw/pal/Cargo.toml
```diff
@@ -46,8 +46,13 @@ embassy-time = { workspace = true }
...
+  "derive",
+  "oid",
+], version = "0.8" }
```
jaygmsft: der crate added directly instead of workspace.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -438,50 +549,73 @@ impl HsmRsa for UnoHsmPal {
...
 
-        self.mod_exp_priv(io, key_size, priv_key, ciphertext, em)
-            .await?;
+        // `em` holds the decrypted encoded message and then the recovered DB —
+        // both secret. The scoped allocator only rewinds its watermark on drop
+        // and never clears freed DMA, so scrub `em` on *every* exit path
+        // (including `?`-propagated errors) before it can be handed to a later
+        // per-IO allocation. The fallible decode runs in an inner block so the
+        // single `em.zeroize()` below covers all of them.
+        let result = async {
```
jaygmsft: from readability perspective, might be better to extract the async block into a async method.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Good catch - it was expressing the same thing twice. The vault operand's exponent slot is a fixed `EXP_WIRE_LEN` (4-byte) field, so the length is a property of the destination, not something the caller needs to pass. Dropped `e_len`: `e` is now converted to its little-endian wire form once, while parsing, and assembly is a single `copy_from_slice` instead of copy + reverse + zero-pad. Also hoisted `EXP_WIRE_LEN` to module scope so the width is not spelled `4` in some places and named in others.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Agreed - removed. Took out every explicit zeroize in `rsa.rs` (this one plus the four in `rsa_priv_der_to_vault`), not just this line, since the same argument applies to all of them. Left a comment at the top of `rsa_priv_der_to_vault` noting the buffer is intentionally left dirty and the per-IO teardown scrub is what wipes it, so it does not read as an oversight.

fw/plat/uno/fw/pal/Cargo.toml
radutta99: Done - moved to `[workspace.dependencies]` in `fw/plat/uno/fw/Cargo.toml` and the pal now uses `der = { workspace = true }`, consistent with the other external crates.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Resolved by deleting it rather than extracting it: the `async` block existed only to funnel every exit path into the single `em.zeroize()`. With that zeroize removed (your comment below), the block had no purpose, so `rsa_oaep_decrypt` is back to a flat `?` flow - no wrapper method needed.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: You were right on both counts, and this is now implemented.

Confirmed the gap: `std` zeroizes on `buf_pool::free` (it fills both the dma and nondma buffers up to their watermarks), while uno only rewound the bump watermark and never wiped. So uno was leaving key material resident across IOs.

Fixed on `user/radutta/io-buf-zeroize-gdma` (pushed, 3 commits), done with the DMA engine as you asked:

- **GDMA-driven wipe.** GDMA has no fill/memset opcode - the SQ entry `OP_CODE` only encodes read-src/write-dst - so the wipe is a DMA copy whose source is the SoC "Dummy Memory" read-as-zero window at `0xA0B0_0000`. That region was not modelled anywhere in the firmware, so it is now an RDL entry (`dummy_mem.rdl`) and the base/size are generated into the reg crate rather than hardcoded. The CPU is only used as a fallback if the GDMA submission fails (logged via the trace facade), and for the DTCM heap, which is not on the GDMA fabric.
- **Every IO.** The scrub hangs off `drop_io`, which the core dispatch loop calls on both the completed and the dropped paths.
- **Internal provisioning too.** The admin slot never passes through `drop_io`, so identity/enable-time keygen and boot-key masking now run inside a `with_admin_io` session that scrubs on exit - obtaining an admin IO *is* entering a scrubbed scope, so no admin path can forget it.
- **Only the dirty region.** Wiping the full 18 KiB per IO was wasteful, so a per-slot high-water mark bounds the wipe to the bytes actually written since the last scrub.

Hardware: the GDMA wipe path is proven on silicon - instrumented firmware confirmed `18432B verified zero via GDMA` with no CPU fallback, and the regression suite (cert-chain incl. multithread, ecc-sign) passed. The final dirty-prefix + admin-session state still needs a re-run; the rig is down right now, so that is queued for tomorrow.

Given that, I have removed the local zeroize calls from this PR per your other comment - the central teardown scrub covers them.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -441,6 +556,13 @@ impl HsmRsa for UnoHsmPal {
         self.mod_exp_priv(io, key_size, priv_key, ciphertext, em)
             .await?;
 
+        // `mod_exp_priv` is little-endian (PKA-native): both the ciphertext
+        // operand and the `em` result are LE wire form. The RSA-OAEP decode
+        // below (RFC 8017 EME-OAEP) operates on the big-endian encoded
+        // message `EM = 0x00 ‖ maskedSeed ‖ maskedDB`, so flip the result
+        // LE→BE first (the std PAL does the same flip around OpenSSL).
```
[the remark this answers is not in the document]
radutta99: Intentional, and covered centrally rather than here.

The explicit `em.zeroize()` was removed at review request: uno now scrubs the whole per-IO DMA/DTCM scratch on IO teardown (`drop_io`), using the GDMA engine, so a per-function wipe is redundant work that would have to be deleted again. `em` is a scoped allocation inside that same per-IO arena, so it is covered on every exit path - including `?`-propagated errors - without needing the wrapper block that previously existed only to funnel exits into a single scrub.

That work is on `user/radutta/io-buf-zeroize-gdma`.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -160,14 +251,38 @@ impl HsmRsa for UnoHsmPal {
...
+        let vault_len = 2 * k + EXP_WIRE_LEN;
+        // In-place assembly stages `d` into `buf[vault_len..]`, so the recovered
+        // DER must be at least `vault_len + d_len` bytes. A full RSAPrivateKey is
+        // always larger, but reject rather than panic on a malformed / truncated
+        // DER.
+        if buf.len() < vault_len + layout.d_len {
+            return Err(HsmError::InvalidArg);
+        }
+        assemble_rsa_operand_in_place(&mut buf[..], &layout);
+        Ok((vault_len, k))
```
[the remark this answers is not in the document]
radutta99: Same as the `rsa_oaep_decrypt` comment: intentional, and handled centrally.

The premise "there is no allocator/IO-slot teardown wipe here" was true when this PR was written, but that is exactly the gap being closed on `user/radutta/io-buf-zeroize-gdma`: the per-IO SRAM and DTCM scratch is now scrubbed on `drop_io` via the GDMA engine, covering both the early-return paths (CRT requested / parse failure / size check) and the DER tail left after a successful import.

`rsa_priv_der_to_vault` carries a comment noting the buffer is deliberately left dirty and what wipes it, so it does not read as an oversight.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -67,6 +68,97 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
...
+    let e = key.public_exponent.as_bytes();
+    let d = key.private_exponent.as_bytes();
+    let modulus_len = n.len();
+    if !matches!(modulus_len, 256 | 384 | 512) || e.len() > EXP_WIRE_LEN || d.len() > modulus_len {
+        return None;
+    }
+    // DER `e` is big-endian and minimally encoded; the operand slot is a fixed
+    // little-endian `EXP_WIRE_LEN` field, so reverse into a zeroed array.
+    let mut e_le = [0u8; EXP_WIRE_LEN];
+    for (dst, &src) in e_le.iter_mut().zip(e.iter().rev()) {
```
jaygmsft: i think there is some reverse copy api already in firmware that you can use for this.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -67,6 +68,97 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
...
+    buf.copy_within(layout.d_off..layout.d_off + d_len, vault_len);
+    // 2. vault `n` at [k, 2k): move DER `n` (big-endian, exactly k bytes) into
+    //    place, then reverse to little-endian.
+    buf.copy_within(layout.n_off..layout.n_off + k, k);
+    buf[k..2 * k].reverse();
+    // 3. vault `d` at [0, k): move the staged big-endian `d` to the front,
+    //    reverse to little-endian, and zero-pad the high bytes.
+    buf.copy_within(vault_len..vault_len + d_len, 0);
+    buf[0..d_len].reverse();
+    buf[d_len..k].fill(0);
```
jaygmsft: i dont think we need to fill with zero. buf should already be all zeros right?

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -67,6 +68,97 @@ fn digest_info_prefix(algo: HsmHashAlgo) -> &'static [u8] {
...
+/// moved and reversed big-endian→little-endian into place, followed by the staged
+/// `d` and the saved `e`. Requires `buf.len() >= vault_len + d_len` (checked by
+/// the caller). The tail is left dirty; the per-IO teardown scrub wipes it.
+fn assemble_rsa_operand_in_place(buf: &mut [u8], layout: &RsaOperandLayout) {
+    let k = layout.modulus_len;
+    let vault_len = 2 * k + EXP_WIRE_LEN;
+    let d_len = layout.d_len;
+    // 1. Stage big-endian `d` into the tail scratch before its source is
+    //    clobbered by the `n` move below.
+    buf.copy_within(layout.d_off..layout.d_off + d_len, vault_len);
```
jaygmsft, line 148: You are doing bunch of copies and then doing reverse in place. lets do reverse copy directly perhaps?

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Good catch - there was one, four times over. `reverse_copy(dst, src)` was implemented privately in the std PAL drivers, in `core/evidence`, in the TBOR `key_report` handler and in the MBOR `attest_key` handler, all with the same body, and this loop was a fifth copy.

Moved it into `azihsm_fw_hsm_pal_traits` (every one of those crates already depends on it) and deleted the four duplicates. The shared version keeps the permissive `zip` behaviour the core copies had, plus the std copy''s `debug_assert` that `dst` is long enough, so a too-short `dst` is caught in debug instead of silently truncating.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Applied where the ranges allow it, but not everywhere - the three fields differ:

- **`e`**: now a `reverse_copy` straight into the layout''s little-endian array at parse time, so assembly is a single `copy_from_slice`.
- **`d`**: the staged copy at `[vault_len..]` is disjoint from the destination `[0, d_len)`, so this is now one `split_at_mut` + `reverse_copy` instead of `copy_within` + in-place `reverse` - one pass instead of two.
- **`n`**: kept as `copy_within` + in-place `reverse`. Its source is the DER modulus at `n_off`, which can overlap the destination `[k, 2k)`, and `reverse_copy` needs two non-overlapping slices. `copy_within` is a memmove, so it stays correct under overlap.

The initial staging of `d` into the tail also stays a `copy_within` for the same overlap reason.

I checked the rewritten assembly is byte-identical to the previous one across 60 combinations of modulus size (2048/3072/4096), `d` length and `n`/`d` offsets.

Same pattern folded into four `copy_from_slice` + `reverse` pairs in the deterministic ECC signing path (`ecc_det.rs`), which all copy between distinct buffers.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: `buf` is not zero here - it holds the recovered plaintext DER. This function rewrites the vault operand *in place over that same buffer*, so everything it has not explicitly written still contains DER bytes.

Tracing it: the modulus write covers `[k, 2k)`, the staging copy covers the tail, and the `d` write covers `[0, d_len)`. Nothing has touched `[d_len, k)`, so it still holds whatever DER byte happened to live at that offset. The vault `d` field is a fixed `k` bytes, so without the fill those high bytes become part of the private exponent and the imported key is wrong.

It only bites when `d_len < k`, i.e. when the DER `INTEGER` for `d` had a leading zero byte that `UintRef` stripped - roughly 1 key in 256, and more for multiple leading zeros. That rarity is what makes it worth keeping: without the fill it would be an intermittent, silent key-corruption bug rather than an obvious one.

(For the same reason `dma_alloc` does not zero - `dma_alloc_zeroed` is the zeroing variant - and even a zeroed buffer would have been overwritten by the DER before we get here.)

I have left the fill in; happy to revisit if you would rather the operand be built into a separate zeroed buffer instead.

jaygmsft (approved):

fw/plat/uno/fw/pal/src/crypto/rsa.rs
radutta99: Correction to my reply above — I changed where this landed, so the previous description is out of date.

I had first put the shared `reverse_copy` in `azihsm_fw_hsm_pal_traits` and deduplicated all four existing copies (std PAL drivers, `core/evidence`, TBOR `key_report`, MBOR `attest_key`). That was the wrong home: `reverse_copy` is a plain function, not part of any PAL contract, so it does not belong in a traits crate.

It now lives in the uno PAL at `fw/plat/uno/fw/pal/src/crypto/mod.rs` as a `pub(crate)` helper, used by `rsa.rs` and `ecc_det.rs`. Those shared crates are reverted and are byte-identical to before.

The trade-off, to be explicit: the std PAL and the three core handlers keep their own private copies, since they cannot reach a uno-local helper. Deduplicating across crates would need a shared utility crate, which felt out of scope for this PR. The commit message records that so it does not read as an oversight later.

What is unchanged from that reply: `reverse_copy` is used for the exponent and the private exponent in the operand assembly (the latter now a single pass via `split_at_mut`), the modulus keeps `copy_within` + in-place reverse because its source and destination can overlap, and the four copy-then-reverse pairs in the deterministic ECC signing path are folded.

Branch was amended and force-pushed, so the intermediate `pal_traits` version is not in the history.

fw/plat/uno/fw/pal/src/crypto/rsa.rs
```diff
@@ -160,14 +256,38 @@ impl HsmRsa for UnoHsmPal {
...
-        // TODO: parse the recovered RSA private-key DER on Uno and rewrite
-        // it in place into the vault representation — the raw non-CRT
-        // (`n`/`e`/`d`) or the custom CRT layout selected by `crt`
-        // (RsaUnwrap RSA import).
-        Err(HsmError::UnsupportedCmd)
+        // Note: `buf` holds recovered plaintext DER (and, after assembly, the
+        // staged `d` plus the leftover CRT components `p`/`q`/`dp`/`dq`/`qinv`)
+        // on every path through this function. It is not scrubbed here — the
+        // per-IO teardown scrub wipes the whole slot, so a local wipe would be
+        // redundant work that has to be removed again.
```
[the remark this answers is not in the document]
radutta99: As mentioned in other comments responses, zeroization of the whole input  dma buffer including the bump allocated one is coming as a separate PR so that we don't have to zeroize it piecemeal.

vsonims (approved):

PR #609, opened 2026-07-29
feat(provider): import RSA keys in CRT form

walterchris:
## Summary

Builds on `user/rajeshgali/rsa-crt-support` (the `RsaCrt` key kind + DDI key class mapping) and adds provider support for importing RSA keys in CRT form, which lets the HSM use the faster CRT private-key operations.

The key form is a caller decision, not something inferred from the key material:

- **`azihsm.key_kind`** selects the form for RSA imports — `RSA` or `RSA-CRT`. Unset means `RSA-CRT`. The same selection applies to both import paths (`azihsm.input_key` and `azihsm.wrapped_key`), so a pre-wrapped blob needs no special handling.
- The parameter is the existing `azihsm.key_kind` already used by the AES skeymgmt (`AES`, `AES-GCM`, `AES-XTS`), not a new one.
- The public half of a CRT pair stays plain RSA — there is no CRT public key — both in the import property lists and when reconstructing properties from masked-key metadata.

Plumbing for `HsmKeyKind::RsaCrt`: exposed through the native API as `AZIHSM_KEY_KIND_RSA_CRT`, dispatched for RSA-AES unwrap and masked-key unmask, accepted by RSA private-key props validation, and recognised on store reload.

## Why the form is selected rather than detected

An earlier revision inspected the PKCS#8 DER and picked the form from whether the CRT components were present. That was dropped (see review discussion): deriving the form from the input silently changes what gets stored, which hides the outcome from the caller and makes a performance-relevant property depend on how a key happened to be encoded. Selecting it outright keeps the stored form predictable, and a key that cannot be stored in CRT form fails loudly instead of being quietly downgraded.

Removing the inspection also removed the hand-rolled ASN.1 walk it needed, which is most of the net line reduction in the last commit.

## Tests

- **API tests** (mock, nextest): unwrap a key pair with the `RsaCrt` kind and verify the private key comes back as `RsaCrt` while the public half stays `Rsa`, both directly after unwrap and after a mask/unmask round trip. The plain-RSA forms were already covered by the existing unwrap/unmask tests.
- **Provider integration test** (lit, `testfiles/rsa/key_kind/`): imports the same key with the parameter unset, with `azihsm.key_kind:RSA`, and as a pre-wrapped blob with `azihsm.key_kind:RSA-CRT`, asserting the stored form via the DDI key type in the masked blob metadata (plaintext by design); signs with a reloaded masked CRT key and verifies against the software public key; rejects an unknown parameter value.
- Full runs: provider CLI suite 30/30, provider C-API suite 46/46, RSA API tests 37/37.

## Note for review

The zero-CRT-components case cannot be exercised end to end. A key synthesised from (n, d) only — e.g. exported from CNG or another HSM — will now be declared CRT and handed to the device that way, since nothing inspects it first. That is the intended trade, but it assumes the device rejects such an import rather than accepting it and computing garbage. The mock cannot answer this: its import path just retags the kind. Worth confirming against firmware before this reaches hardware.

Two smaller consequences of defaulting to CRT, both intentional but worth being explicit about: it changes the stored form for anyone importing without the parameter today, and the sim's device-side sizes put a CRT 2K private key at 1284 bytes against 516 for plain (3K: 1924/772, 4K: 2564/1028), so per-key slot usage grows.


plugins/ossl_prov/inc/azihsm_ossl_pkey_param.h
```diff
@@ -23,6 +23,11 @@ extern "C"
 /* Opaque AES key kind for SKEYMGMT import/generate: "AES" (default), "AES-GCM", or "AES-XTS". */
 #define AZIHSM_OSSL_PKEY_PARAM_KEY_KIND "azihsm.key_kind"
 
+/* RSA key form for import: "rsa" or "rsa-crt".  When unset, azihsm.input_key
+ * imports auto-detect the form from the key's CRT components and
+ * azihsm.wrapped_key imports default to plain "rsa". */
+#define AZIHSM_OSSL_PKEY_PARAM_KEY_TYPE "azihsm.key_type"
```
rajesh-gali: do we need new key type? can we use KEY_KIND?

plugins/ossl_prov/src/azihsm_ossl_hsm.c
```diff
@@ -1433,6 +1434,145 @@ static azihsm_status wrap_and_unwrap_pkcs8(
...
+}
+
+/*
+ * Determine whether a PKCS#8-encoded RSAPrivateKey carries usable CRT
+ * components. PKCS#1 makes prime1..coefficient mandatory fields, but keys
+ * synthesized from (n, d) only (e.g. exported from another HSM or CNG)
+ * encode them as zero-valued INTEGERs; such keys can only be imported in
+ * plain RSA form.
+ */
+static bool pkcs8_rsa_has_crt_components(const uint8_t *der, long der_len)
```
rajesh-gali: We can avoid checking for crt components.  Default to CRT enabled if user doesn't specify key kind.

plugins/ossl_prov/inc/azihsm_ossl_pkey_param.h
walterchris: Done — dropped the separate parameter and reused `azihsm.key_kind` for RSA imports, with values `RSA` and `RSA-CRT` to match the AES naming. Fixed in ffed5b37.

plugins/ossl_prov/src/azihsm_ossl_hsm.c
walterchris: Done — removed the CRT-component check entirely. The form is now taken from `azihsm.key_kind` and defaults to RSA-CRT when unset, for both the plaintext and pre-wrapped import paths. Fixed in ffed5b37.

plugins/ossl_prov/src/azihsm_ossl_hsm.c
```diff
@@ -1433,6 +1434,145 @@ static azihsm_status wrap_and_unwrap_pkcs8(
...
+    int tag = 0, cls = 0;
+
+    /* ASN1_get_object() sets 0x80 in the return value on parse errors */
+    if ((ASN1_get_object(pp, out_len, &tag, &cls, max_len) & 0x80) || tag != expected_tag)
+    {
+        return OSSL_FAILURE;
+    }
+
+    return OSSL_SUCCESS;
+}
```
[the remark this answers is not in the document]
walterchris: Obsolete — the DER walk was removed in ffed5b37. The key form is now selected via `azihsm.key_kind` instead of being inferred from the key material, so there is no ASN.1 parsing left in this path.

plugins/ossl_prov/src/azihsm_ossl_hsm.c
```diff
@@ -1433,6 +1434,145 @@ static azihsm_status wrap_and_unwrap_pkcs8(
...
+    const uint8_t *end = der + der_len;
+    long len = 0;
+    int i;
+
+    /* PrivateKeyInfo ::= SEQUENCE { version, privateKeyAlgorithm, privateKey } */
+    if (der_expect(&p, end - p, V_ASN1_SEQUENCE, &len) != OSSL_SUCCESS)
+    {
+        return false;
+    }
+
```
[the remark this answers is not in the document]
walterchris: Obsolete — same as above; the PKCS#8 walk was removed in ffed5b37 along with the auto-detection it supported.

api/lib/src/algo/rsa/key.rs
```diff
@@ -54,7 +54,7 @@ impl HsmRsaPrivateKey {
         let supported_flags = HsmKeyFlags::DECRYPT | HsmKeyFlags::SIGN | HsmKeyFlags::UNWRAP;
 
         // Kind/class: ensure we're validating an AES *secret* key.
```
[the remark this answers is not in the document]
walterchris: Fixed in ffed5b37 — the comment now reads "ensure we're validating an RSA *private* key".

plugins/ossl_prov/integration-tests/openssl-cli/testfiles/rsa/key_kind/key_kind.sh
```diff
@@ -0,0 +1,142 @@
...
+# padding, then the MBOR-encoded metadata whose second field (bytes 15..18,
+# big-endian u32) is the DDI key type: 1 = RSA 2K private, 4 = RSA 2K
+# private CRT.
+masked_key_type() {
+    local iv_len post_iv md_start
+    iv_len=$(od -An -tu2 -j4 -N2 "$1" | tr -d ' ')
+    post_iv=$(od -An -tu2 -j6 -N2 "$1" | tr -d ' ')
+    md_start=$((4 + 48 + iv_len + post_iv))
+    od -An -tu1 -j$((md_start + 18)) -N1 "$1" | tr -d ' '
+}
```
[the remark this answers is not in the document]
walterchris: Fixed in ffed5b37 — the helper now reads all four bytes and combines them as a big-endian u32, matching the comment. (The script also moved to `testfiles/rsa/key_kind/key_kind.sh` with the parameter rename.)

plugins/ossl_prov/src/azihsm_ossl_keymgmt_rsa.c
```diff
@@ -521,6 +560,25 @@ static int azihsm_ossl_keymgmt_gen_set_params(AZIHSM_RSA_GEN_CTX *genctx, const
...
+        if (!azihsm_rsa_kind_from_str(p->data, &genctx->key_kind))
+        {
+            ERR_raise_data(
+                ERR_LIB_PROV,
+                PROV_R_INVALID_KEY,
+                "azihsm: azihsm.key_kind must be 'RSA' or 'RSA-CRT'"
+            );
+            return OSSL_FAILURE;
+        }
+    }
```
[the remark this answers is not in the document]
walterchris: Good catch, and it applied to more than this line — every string parameter in `azihsm_ossl_keymgmt_gen_set_params` (`key_usage`, `session`, `masked_key`, `input_key`, `wrapped_key`, `key_kind`) was passing `p->data` straight to a string function. Fixed all six in bd953be7 via a new shared `azihsm_ossl_get_str_param()` in `azihsm_ossl_pkey_param.c`, which copies into a caller-supplied buffer bounded by the declared length.

Two deviations from the suggestion, both found by the test suites:

1. `OSSL_PARAM_get_utf8_string()` itself was too strict. The `OSSL_PARAM_utf8_string()` initializer macro leaves `data_size` at 0 for a C string (unlike `OSSL_PARAM_construct_utf8_string()`, which fills in `strlen`), and our C-API tests use that macro — so the strict call copied zero bytes and the imports failed. The helper now bounds the scan by `data_size` when non-zero and by the destination size otherwise, which removes the unbounded read without breaking those callers.

2. An absent parameter must leave the destination untouched. OpenSSL calls `gen_set_params` once per `-pkeyopt`, so clearing on absence wiped values stored by earlier calls.

Verified: provider CLI 30/30, C-API 46/46.

api/lib/src/ddi/masked_key.rs
```diff
@@ -330,9 +330,17 @@ impl HsmMaskedKey {
         }
         flags |= HsmKeyFlags::EXTRACTABLE;
 
+        // Only the private half of an RSA pair carries the CRT form; the
+        // public key is always plain RSA.
+        let kind = if class == HsmKeyClass::Public && metadata.kind == HsmKeyKind::RsaCrt {
```
rajesh-gali: this can be simplify using match statement. Allows us to add additional kids easily.

plugins/ossl_prov/src/azihsm_ossl_store.c
```diff
@@ -380,7 +380,7 @@ static int load_and_unmask_key(AZIHSM_STORE_CTX *ctx)
         }
     }
     /* For RSA keys, query bit length and session flag */
-    else if (actual_kind == AZIHSM_KEY_KIND_RSA)
+    else if (actual_kind == AZIHSM_KEY_KIND_RSA || actual_kind == AZIHSM_KEY_KIND_RSA_CRT)
```
rajesh-gali, line 383: should we also have else sectiopn, to return error in case actual_kid is unsupported

api/lib/src/ddi/masked_key.rs
walterchris: Done in 210e2ae4 — it is now a match on the `(class, kind)` pair, so another class-specific kind is a new arm rather than another condition:

```rust
let kind = match (class, metadata.kind) {
    (HsmKeyClass::Public, HsmKeyKind::RsaCrt) => HsmKeyKind::Rsa,
    (_, kind) => kind,
};
```

plugins/ossl_prov/src/azihsm_ossl_store.c
walterchris: Agreed, added in 210e2ae4. The fall-through was returning success with `ec_curve`/`rsa_bits` never populated, and leaving the handles `azihsm_key_unmask_pair` had just created. The else now raises `PROV_R_INVALID_KEY` with the offending kind, calls `store_ctx_delete_key_handles()`, and returns failure.

Worth noting it changes no working path: AES is the only other kind that can reach here, and `store_type_to_ossl_name()` already returns NULL for it, so an `azihsm://…;type=aes` URI failed a few lines later anyway. This just fails where the cause is known and releases the handles on the way out.

PR #619, opened 2026-08-03
fix(capi-tests): select TPM sources via AZIHSM_USE_TPM

milan-zivadinovic:
Update the OpenSSL C API integration test configuration to use TPM-backed OBK and POTA sources when `AZIHSM_USE_TPM` is set.

Caller-provided sources remain the default when the variable is unset.

### DIFF
--- a/plugins/ossl_prov/integration-tests/openssl-capi/cpp/openssl_capi_integration_tests.rs
+++ b/plugins/ossl_prov/integration-tests/openssl-capi/cpp/openssl_capi_integration_tests.rs
@@ -299,8 +299,9 @@ mod integration {
         let provider_so = provider_path.join("azihsm_provider.so");
         let conf_path = keymat_dir.join("openssl.cnf");
 
-        // TPM sources when AZIHSM_USE_TPM is set, caller sources otherwise.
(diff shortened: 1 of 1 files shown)


plugins/ossl_prov/integration-tests/openssl-capi/cpp/openssl_capi_integration_tests.rs
```diff
@@ -299,6 +299,13 @@ mod integration {
         let provider_so = provider_path.join("azihsm_provider.so");
         let conf_path = keymat_dir.join("openssl.cnf");
 
+        // TPM sources when AZIHSM_USE_TPM is set, caller sources otherwise.
+        let (obk_source, pota_source) = if env::var("AZIHSM_USE_TPM").is_ok() {
```
rajesh-gali: This will cause obk_soure, pota_source to use TPM, though user sets ENV as AZIHSM_USE_TPM=0

milan-zivadinovic: Fixed. Please check.

rajesh-gali (approved):

PR #625, opened 2026-08-05
fw(uno):Added mcr_test_hooks and fips_validation_hooks feature flags

aminkrutik-micro:
Adds two Cargo feature flags to the refactored HSM firmware so upcoming
test/validation-only ops can be gated behind them:

- mcr_test_hooks       — for test-hook ops (e.g. der_key_import)
- fips_validation_hooks — for FIPS validation ops (e.g. raw_key_import, get_priv_key)

### DIFF
--- a/fw/core/lib/Cargo.toml
+++ b/fw/core/lib/Cargo.toml
@@ -36,5 +36,7 @@ zerocopy.workspace = true
 
 [features]
 default = ["std"]
+fips_validation_hooks = []
+mcr_test_hooks = []
 std = []
 tracing = []
--- a/fw/plat/std/lib/Cargo.toml
+++ b/fw/plat/std/lib/Cargo.toml
@@ -5,6 +5,10 @@ edition = "2021"
 name = "azihsm_fw_hsm_std"
 version = "0.1.0"
 
+[features]
+fips_validation_hooks = ["azihsm_fw_hsm_core/fips_validation_hooks"]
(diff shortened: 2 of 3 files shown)


fw/core/lib/Cargo.toml
```diff
@@ -36,5 +36,7 @@ zerocopy.workspace = true
 
 [features]
 default = ["std"]
+fips_validation_hooks = []
+mcr_test_hooks = []
```
vsonims: We should not be modifying core for test hooks.

vsonims (changes requested):

fw/core/lib/Cargo.toml
aminkrutik-micro: Agreed, core should stay free of test scaffolding. But where we put the feature flag is tied to where the handlers live, so I want to align on direction first.

**Current state**: DDI dispatch is a single match hdr.op  in `fw/core/lib/src/ddi/mbor/mod.rs`, and every handler is a pub(crate) mod inside core. There is no extension point today, so a test-hook op (`raw_key_import`, `der_key_import`, `get_priv_key`, `test_action`) needs its handler and its match arm in core — which means the #[cfg(feature)] gate, and the feature definition, must also be in core. The flags in this PR are inert ( [] , off by default, no effect on production builds).

Which direction do we should follow?
- Keep the gated ops in core behind these off-by-default features (current approach) — a feature flag in core, but no production code path.

Please suggest what you think.

PR #630, opened 2026-08-05
ci: harden apt against Ubuntu mirror outages

walterchris:
archive.ubuntu.com resolves round-robin across several IPs, and a single dead one fails a matrix cell. Worse, apt-get update exits 0 when a mirror is unreachable, so the job died later in apt-get install with a misleading "package has no installation candidate" instead of a network error.

For the debian-family container cells:
- retry transient fetches (Acquire::Retries 3)
- run apt-get update with --error-on=any so an unreachable mirror fails the step at the point of failure

The images' stock sources.list is left untouched: archive.ubuntu.com stays the mirror. --error-on=any is supported by every apt in the matrix, oldest being 20.04's 2.0.10.

### DIFF
--- a/.github/workflows/engine-matrix.yml
+++ b/.github/workflows/engine-matrix.yml
@@ -66,13 +66,8 @@ jobs:
         env:
           DEBIAN_FRONTEND: noninteractive
         run: |
-          # Use the Azure mirror: archive.ubuntu.com is round-robin DNS where
-          # one dead IP fails the job.  No-op on Debian (deb.debian.org).
-          for f in /etc/apt/sources.list /etc/apt/sources.list.d/*.sources; do
-            [ -f "$f" ] && sed -i \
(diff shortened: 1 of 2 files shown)


.github/workflows/engine-matrix.yml
```diff
@@ -66,7 +66,17 @@ jobs:
         env:
           DEBIAN_FRONTEND: noninteractive
         run: |
-          apt-get update
+          # Use the Azure mirror: archive.ubuntu.com is round-robin DNS where
+          # one dead IP fails the job.  No-op on Debian (deb.debian.org).
+          for f in /etc/apt/sources.list /etc/apt/sources.list.d/*.sources; do
+            [ -f "$f" ] && sed -i \
+              -e 's|http://archive.ubuntu.com|http://azure.archive.ubuntu.com|g' \
```
rajesh-gali, line 73: do we need azure.archive? I think retrying 3 times is enough without touching the system files.

jaygmsft (approved):

PR #632, opened 2026-08-06
uno: zeroize per-IO DMA buffers on teardown using the GDMA engine

radutta99:
## What

Uno never wiped its per-IO scratch. `std` zeroizes on `buf_pool::free`, but uno only rewound the bump watermark, so key material — RSA private keys, the P-384 identity scalar, OAEP plaintext — stayed resident in the per-IO SRAM/DTCM heaps and was visible to whichever IO reused the slot. Uno has no heap, so those bump buffers are the *only* place such material exists.

This adds a teardown scrub driven by the GDMA engine, per @jaygmsft's review comment on #602.

## How

**The wipe is a DMA copy from a read-as-zero window.** GDMA has no fill/memset opcode — an SQ entry's `OP_CODE` only encodes read-src / write-dst — so zeroing means copying out of the SoC's "Dummy Memory" region at `0xA0B0_0000`. That region wasn't modelled anywhere in the firmware, so it is now an RDL entry (`dummy_mem.rdl`) with the base and size generated into the reg crate rather than hardcoded.

**Every host IO.** The scrub hangs off `drop_io`, which the core dispatch loop calls on both the completed and the dropped paths.

**Internal provisioning too.** The admin slot never passes through `drop_io`, so identity/enable-time keygen and boot-key masking now run inside a `with_admin_io` session that scrubs on exit — obtaining an admin IO *is* entering a scrubbed scope. `UnoHsmIo::admin_no_scrub` is the one deliberate opt-out, named so a new call site has to choose it.

**Only what was written.** A per-(slot, heap) high-water mark bounds the wipe. `dma_alloc_var*` reserve the whole remaining heap so the callback need not know its encoded length up front; they snapshot the peak before reserving and restore it to the trimmed length afterwards, and document the contract that rests on. On the error path the peak stays at the full reservation, since a partial write before a failure is unbounded.

## Hardware validation

Flashed to the Manticore EVB and instrumented to scan each slot's **whole capacity** after every scrub, classifying any non-zero byte as inside or beyond the wiped region:

- **No byte past the recorded peak, ever** — across ~1.7k IOs on every slot. This is the premise of wiping a prefix, and it also confirms the `dma_alloc_var` contract holds for both the MBOR encoder and the TBOR frame builders.
- **No residue inside the wiped region**, and **no CPU fallback** — GDMA handled every wipe.
- One representative line: a slot wiped only 288 B, and the full 18432 + 1536 B read back zero.

Cost: the DMA peak averages **~933 B of the 18 KiB slot**, max **~3.3 KiB** — so the wipe fits in a single `DUMMY_MEM`-sized GDMA transfer instead of two.

Regression: 45/45 over three passes (cert-chain incl. multithread, ECC-sign compat across P-256/384/521, sign/verify, part-info). `rsa_unwrap` and `attest_key` match this branch's baseline exactly; their failures are pre-existing bring-up gaps (`UnsupportedCmd` for the PKA-stubbed ECC/RSA attestation, `FileHandleNoExistingSession` for NoSession routing).

## Notes for review

- `io_peak` starts at zero, so each slot's first scrub covers only what that boot wrote. That relies on the IO heaps arriving clean, which 1SP guarantees on a warm boot; the field doc records the dependency and what to change if that ever stops holding.
- The DTCM heap is not on the GDMA fabric, so it is wiped by the CPU — word-wide, and with volatile writes that cannot be elided.
- A GDMA failure falls back to a CPU wipe and is logged through the trace facade, so a persistent fault cannot silently degrade into an unsignalled byte loop.

## Merge order

#602 removes its local zeroize calls on the assumption this lands, so if #602 merges first there is a window with no wipe on those paths. Worth sequencing this one first.


fw/plat/uno/fw/pal/src/alloc.rs
```diff
@@ -414,15 +495,27 @@ impl HsmAlloc for UnoHsmPal {
...
+        let peak_before = pk(self, io_index, DMA).with(|v| *v);
         let (start, buf) = bump(self, io_index, DMA, cap - aligned, 4)?;
         match f(buf) {
             Ok(len) => {
-                w.with(|v| *v = start + len.min(buf.len()));
+                let end = start + len.min(buf.len());
+                w.with(|v| *v = end);
+                pk(self, io_index, DMA).with(|v| *v = peak_before.max(end));
                 // SAFETY: `buf` came from the SRAM Dma pool.
                 Ok(unsafe { DmaBuf::from_raw_mut(&mut buf[..len]) })
```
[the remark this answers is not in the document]
radutta99: Good catch — fixed in 819d6136.

The two really did disagree: `end` was clamped with `len.min(buf.len())` but the slice was not, so an over-reported `len` would panic and take the firmware down rather than failing the one command.

Rather than clamping, I matched what the std PAL already does for the same callback contract (`fw/plat/std/pal/src/alloc.rs`): reject with `InvalidArg` and refuse to hand back a longer slice than we own. Silently truncating would turn an encoder bug into a short or malformed response the host has to interpret, which seems worse than failing the command outright — and it keeps the two PALs behaving the same way.

One detail specific to this PR: on that rejection the watermark rewinds but the *peak* is deliberately left at the full reservation. A callback that over-reports wrote an unknown amount, so the teardown scrub still has to cover the whole reserved span. Same reasoning as the existing error path.

The `# Contract` doc now states the rejection behaviour too.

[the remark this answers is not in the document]
radutta99: Follow-up: the status is now `DmaAllocLenOverrun` rather than `InvalidArg` (c14ce4bc).

`InvalidArg` was wrong here by the enum's own convention — its docs reserve it for a caller bug ("unknown id, kind mismatch", "a malformed request"), but in this case the host's request is perfectly well-formed and the fault is ours: one of our encoders mis-reporting its output length. Returning `InvalidArg` would blame the host for a firmware bug.

Added a dedicated status following the `UndoLogFull` precedent, which documents exactly this shape of problem — a should-never-happen firmware bug that gets its own code so it is diagnosable rather than hidden behind a generic one. The doc spells out how it differs from `InvalidArg` (malformed request) and `NotEnoughSpace` (genuine exhaustion).

Both PALs now return it for the same condition, so std and uno stay consistent.

fw/plat/uno/fw/pal/src/gdma.rs
```diff
@@ -218,3 +234,116 @@ impl HsmGdmaController for UnoHsmPal {
...
+    /// GDMA-zero a GDMA-reachable device region `[dst_ptr, dst_ptr + len)`.
+    ///
+    /// Copies zeros from the [`DUMMY_MEM_BASE`] window, chunked to its
+    /// [`DUMMY_MEM_SIZE`] so an arbitrarily large region can be wiped
+    /// with a fixed 16 KiB zero source. Both operands use the device
+    /// interface. Only valid for GDMA-reachable memory (GSRAM); the M7 TCM
+    /// is not on the GDMA fabric and must be wiped by the CPU.
+    async fn gdma_zero_region(&self, dst_ptr: *mut u8, len: usize) -> HsmResult<()> {
+        let mut off = 0usize;
+        while off < len {
```
jaygmsft, line 249: instead of while loop submitting multiple DMA requests, cant we zeroize a specific buffer in a single command?

jaygmsft: We know that this buffer will never be bigger than 8KB

fw/plat/uno/fw/pal/src/gdma.rs
```diff
@@ -218,3 +234,116 @@ impl HsmGdmaController for UnoHsmPal {
...
+/// wipe on the IO teardown path.
+///
+/// # Safety
+///
+/// `ptr` must be valid for writes of `len` bytes.
+#[inline]
+unsafe fn cpu_zeroize(ptr: *mut u8, len: usize) {
+    let mut i = 0usize;
+    // Word body: 4 bytes per store (both per-IO regions hit this path).
+    while i + 4 <= len && (ptr as usize + i).is_multiple_of(4) {
```
jaygmsft: I think ARM can do QWORD as well. Also doesnt zeroize crate have a method for this that we can call?

fw/plat/uno/fw/pal/src/io.rs
```diff
@@ -18,13 +18,14 @@
 //! | `IO_CQ[index]`  | 16B CQE           | Completion queue entry (write)       |
 //! | `IO_META[index]` | 8B metadata      | Controller/queue IDs from IIC recv   |
 //! | `DTCM_IO_BUF[index]` | 1.5KB fmem   | Fast DTCM workspace buffer           |
-//! | `SRAM_IO_BUF[index]` | 8KB smem     | Large SRAM workspace buffer          |
+//! | `SRAM_IO_BUF[index]` | 18KB smem    | Large SRAM workspace buffer          |
```
jaygmsft, line 21: Did you verify that the 18KB is in error and should be 8KB instead?

fw/plat/uno/fw/pal/src/gdma.rs
radutta99: Even if the IO buffer shrinks to 8 KB the loop is probably needed. it decouples the wipe from both the destination size and the source window size.  We can choose to keep 8KB as the limit but if dummy memory shrinks from 16KB to less, we may need the loop then right?

fw/plat/uno/fw/pal/src/gdma.rs
radutta99: Correction to my earlier note — I tried it and the `zeroize` crate **does** get us the QWORD, so I've switched to it (7dbbe273).

Zeroizing a `&mut [u64]` view of the region (crate impl: `impl<Z: DefaultIsZeroes> Zeroize for [Z]`, and `u64: DefaultIsZeroes`) lowers to 64-bit `STRD` on Cortex-M7. Verified in the release disassembly — the wipe loop is a 4x-unrolled doubleword store of a zero pair:

```
movs r6, #0
strd r6, r6, [r0, #32]
strd r6, r6, [r0, #40]
strd r6, r6, [r0, #48]
strd r6, r6, [r0, #56]
bne  <loop>
```

So `cpu_zeroize` is now: an 8-byte-aligned `u64` (QWORD) body via the crate, framed by a byte head/tail for any unaligned edges — both per-IO regions are 8-aligned 8-multiples so it's a pure `u64` wipe. Half the stores of the old 32-bit loop, and via an audited primitive rather than a hand-rolled one. The crate still emits the volatile writes + fence so the wipe can't be elided. (This only affects the CPU fallback + the 1.5 KiB DTCM path; the 18 KiB SRAM slot is still GDMA.)

fw/plat/uno/fw/pal/src/io.rs
radutta99: Will wait for @vsonims on this question. On my local testing with enabled DDI tests, 8KB uno works just fine. 
Also note that as per the changes in this PR, zeroization performance does NOT depend on this size. It depends on how much size the IO end up making dirty in its lifetime. So we can leave the size as is today with no efficiency loss.

fw/plat/uno/rdl/soc/dummy_peripheral.rdl
```diff
@@ -0,0 +1,51 @@
...
+//
+// The 16 KiB below is the decoded width of the AXI window as published
+// in the SoC map, not a buffer anyone has to provide or keep free.
+
+
+// ══════════════════════════════════════════════════════════════
+// Read-as-zero window
+// ══════════════════════════════════════════════════════════════
+
+mem dummy_peripheral_t {
```
vsonims: rename to zero peripheral

fw/plat/uno/rdl/soc/uno.rdl
```diff
@@ -51,6 +53,7 @@ addrmap uno {
     aes    AES             @ 0xA020_0000;
     rng    RNG             @ 0xA030_0000;
     sha    SHA             @ 0xA040_0000;
+    dummy_peripheral DUMMY_PERIPHERAL @ 0xA0B0_0000;
```
vsonims: zero ZERO @ 0xA0B0_0000

vsonims (changes requested):

fw/plat/uno/rdl/soc/uno.rdl
radutta99: renamed.

fw/plat/uno/rdl/soc/dummy_peripheral.rdl
radutta99: renamed.

vsonims (approved):

PR #646, opened 2026-08-12
increate  tests coverage for fw_error_decode

meilu-git:


ddi/tbor/types/tests/commands/forward_compat.rs.bak
rajesh-gali: *.bak file shouldn't be checked in.

rajesh-gali (changes requested):

ddi/tbor/types/tests/commands/forward_compat.rs.bak
meilu-git: removed

rajesh-gali (approved):

PR #667, opened 2026-08-21
ECDH derive through the engine's EC pkey method

walterchris:
> **Stacked on #661** (`engine/key-delete`) so this diff shows only the ECDH changes; GitHub retargets to `main` automatically when #661 merges.

## What

ECDH key derivation for the engine, completing the `keyAgreement` half of the EC story: `openssl pkeyutl -derive` and the `EVP_PKEY_derive` C ABI now perform ECDH on the HSM for HSM-backed local keys, and `azihsm.key_usage:keyAgreement` generates derive-capable key pairs instead of being rejected.

## Design

The shared secret is private key material and never leaves the HSM. As in the 3.x provider's keyexch, the output of a derive is the **masked blob of the derived shared-secret key**:

- **Buffer mode** (default): the blob is copied into the caller's buffer; the size query reports the provider's `MASKED_KEY_MAX_BUFFER` (8192) since the blob length is unknown until the derive runs.
- **File mode**: the provider's `output_file` parameter (same un-prefixed name, so `pkeyutl` invocations port between the plugins) writes the blob via the engine's hardened secret-file path; the size query reports 1 and no bytes are returned.

Dispatch follows the established ownership rules: the derive override runs only when the local key carries an attached HSM key; software local keys delegate wholesale to the built-in derive (size query included), pinned byte-for-byte by a unit test. A delegated derive armed with `output_file` fails loudly rather than silently ignoring the option. The peer is an ordinary software public key, validated for group match at `EVP_PKEY_derive_set_peer` and re-validated by the API layer; the ephemeral derived handle is deleted (best-effort, via #661's helper) immediately after masking.

`keyAgreement` keygen sets `can_derive` on both halves (the API enforces sign/derive exclusivity), so generated keys derive but refuse to sign; `azihsm.session` keeps its clear rejection.

## Testing

- **Unit (mock)**: per-curve round trips — generate `keyAgreement` → derive in both modes → reload the masked EC blob and derive through the loaded key; agreement keys refuse to sign; curve-mismatch peers rejected at set_peer; software-key delegation agrees in both directions; `output_file` rejection on delegated derives.
- **CLI**: lit suite per curve — `genpkey` with explicit `azihsm.key_usage:keyAgreement`, then `pkeyutl -derive` with the blob loaded via `azihsm://`, in `output_file` and buffer modes.
- **C ABI**: keygen with the engine handle released immediately after context creation, then derives through contexts created **without** an engine handle — pinning that the ameth-bound key resolves the engine by itself — with both output modes' size-query constants asserted.
- **Hardware**: `derive_ec_key_from_env_smoke` mirrors the mock round trip against a real device (same `AZIHSM_*` env setup as the existing keygen/sign smokes):

  ```
  cargo test -p azihsm_ossl_engine --features engine derive_ec_key_from_env_smoke -- --ignored --nocapture
  ```

All verified in the Ubuntu 20.04 + OpenSSL 1.1.1f container: build, clippy (`-D warnings`, with and without mock), 38+30 unit tests, 4/4 CLI and 4/4 C-ABI integration tests.

## Notes for review

- A masked shared-secret blob is a *generic secret* key, not an EC key: it is not loadable through the EC loader; its consumers are HSM-side KDF/MAC operations, which are follow-up work. At this layer the verifiable contract is blob production and the negative paths — Z-correctness itself is pinned by the api/ddi ECDH tests.
- lit-1.0.4 parses any uppercase word followed by a colon in script comments as a directive — this bit the derive.sh header during development, which now carries a short warning comment about it.


plugins/ossl_engine/integration-tests/openssl-capi/src/engine_capi_tests.rs
```diff
@@ -409,3 +409,117 @@ fn keygen_ec_key_via_engine_capi() {
...
+fn derive_ec_key_via_engine_capi() {
+    let engine_so = std::env::var("ENGINE_SO").expect("ENGINE_SO must point to the engine .so");
+
+    let dir = setup_keymat();
+    let blob = dir.join("agree_ec_key.bin");
+    let out = dir.join("derived_secret.bin");
+
+    let cstr = |s: &str| CString::new(s).unwrap();
+    let curve_key = cstr("ec_paramgen_curve");
+    let curve_val = cstr("P-384");
```
rajesh-gali: can we add test to check all p256, p512 too?

rajesh-gali (approved):

